PR #15, opened 2017-03-14
Windows ansible provision support

pomadchin:
As Windows has no native Ansible support, here is a convenient solution how to achieve provisioning  on Windows. Solution is inspired by https://github.com/ontic/ansible-windows, and uses VM to provision everything.


lossyrob:
@pomadchin what's going on with this PR?

lossyrob:
@pomadchin hey, can you explain what's up with this commit? Were you running this on windows?

pomadchin:
Ah, sorry for that, updated the description; Yes, it allows to run and provision Vagrant on Windows.

lewfish:
I merged a PR that contains this commit. See #86

PR #16, opened 2017-04-19
Revert WIP commits that update to keras 2 from develop

lossyrob:
There's issues with the update to Keras 2 and tensorflow 1. This commit undoes the changes from the develop branch so that it remains workable until we can sort out the problems and do a proper upgrade.


lossyrob:
__Note:__ The "Update Vaihingen" commit needs to be done regardless of the keras version, so can be an independent update.

PR #39, opened 2017-06-08
Changes to normalization procedures

citerana:
We decided to add the ability to calculate the normalization parameters during the preprocessing stage on a sample size of 1000 images and then save the mean and standard deviations to a json file. Once stored, these values are read for future runs. This helps reduce randomness between instantiations of image generators.

You can test the functionality of the parameter saving feature on any image generator by plotting and checking the debug plots or preprocessing the data and checking the json file. For example, the command `python -m rastervision.semseg.data.factory isprs/potsdam numpy preprocess` may be used to store the parameters for the `PotsdamNumpyImageGenerator`. Additionally, unittests were added to ensure consistent normalization behavior. Using the command `python -m rastervision.semseg.data.test.test_batch_norms` while in the Vagrant environment will run this test.

### DIFF
--- a/src/rastervision/common/data/generators.py
+++ b/src/rastervision/common/data/generators.py
@@ -1,8 +1,11 @@
+from os.path import exists, join
+import json
+
 import numpy as np
 from skimage import transform
 
-from rastervision.common.utils import get_channel_stats
 from rastervision.common.settings import TRAIN, VALIDATION, TEST
+from rastervision.common.utils import get_channel_stats, save_json
 
 ROTATE90 = 'rotate90'
 HFLIP = 'hflip'
@@ -74,11 +77,18 @@ class FileGenerator(Generator):
         if self.cross_validation is not None:
             self.process_cross_validation()
 
-        gen = self.make_split_generator(
-            TRAIN, target_size=(10, 10), batch_size=100, shuffle=True,
-            augment_methods=None, normalize=False, only_xy=False)
-        batch = next(gen)
-        self.normalize_params = get_channel_stats(batch.all_x)
+        # If a dataset's normalized parameters have already been
+        # calculated, load its json file. Otherwise, calculate parameters
+        # with a small batch.
+        channel_stats_path = join(self.dataset_path,
+                                  self.name + '_channel_stats.json')
+        if exists(channel_stats_path):
+            with open(channel_stats_path) as channel_stats_file:
+                channel_stats = json.load(channel_stats_file)
+            self.channel_stats = (np.array(channel_stats['means']),
+                                     np.array(channel_stats['stds']))
+        else:
+            self.channel_stats = self.compute_channel_stats(100, False)
 
     def calibrate_image(self, normalized_image):
         calibrated_image = normalized_image.copy()
@@ -193,13 +203,13 @@ class FileGenerator(Generator):
             yield batch, batch_file_inds
 
     def normalize(self, batch_x):
-        means, stds = self.normalize_params
+        means, stds = self.channel_stats
         batch_x = batch_x - means[np.newaxis, np.newaxis, np.newaxis, :]
         batch_x = batch_x / stds[np.newaxis, np.newaxis, np.newaxis, :]
         return batch_x
 
     def unnormalize(self, batch_x):
-        means, stds = self.normalize_params
+        means, stds = self.channel_stats
         nb_dims = len(batch_x.shape)
         if nb_dims == 3:
             batch_x = np.expand_dims(batch_x, 0)
@@ -292,3 +302,17 @@ class FileGenerator(Generator):
         split_gen = map(transform_img_batch, img_batch_gen)
 
         return split_gen
+
+    def compute_channel_stats(self, batch_size, normalize):
+        gen = self.make_split_generator(
+            TRAIN, target_size=(10, 10), batch_size=batch_size,
+            shuffle=True, augment_methods=None, normalize=normalize,
+            only_xy=False)
+        batch = next(gen)
+        return get_channel_stats(batch.all_x)
+
+    def write_channel_stats(self, datasets_path):
+        means, stds = self.compute_channel_stats(1000, False)
+        param_dict = {'means': means.tolist(), 'stds': stds.tolist()}
+        param_path = join(datasets_path, self.name + '_channel_stats.json')
+        save_json(param_dict, param_path)
--- a/src/rastervision/semseg/data/potsdam.py
+++ b/src/rastervision/semseg/data/potsdam.py
@@ -74,6 +74,7 @@ class PotsdamImageFileGenerator(PotsdamFileGenerator):
     def __init__(self, datasets_path, active_input_inds,
                  train_ratio=0.8, cross_validation=None):
         self.dataset_path = join(datasets_path, POTSDAM)
+        self.name = "potsdam_image"
         super().__init__(active_input_inds, train_ratio, cross_validation)
 
     @staticmethod
@@ -90,6 +91,9 @@ class PotsdamImageFileGenerator(PotsdamFileGenerator):
             im_fix = np.zeros((6000, 6000), dtype=np.uint8)
             im_fix[:, 0:-1] = im[:, :, 0]
             save_img(im_fix, file_path)
+
+        PotsdamImageFileGenerator(
+            datasets_path, [0, 1, 2, 3, 4]).write_channel_stats(data_path)
 
     def get_file_size(self, file_ind):
         ind0, ind1 = file_ind
@@ -153,6 +157,7 @@ class PotsdamNumpyFileGenerator(PotsdamFileGenerator):
         self.raw_dataset_path = join(datasets_path, POTSDAM)
         self.dataset_path = join(datasets_path, PROCESSED_POTSDAM)
         self.download_dataset(['processed_potsdam.zip'])
+        self.name = "potsdam_numpy"
 
         super().__init__(active_input_inds, train_ratio, cross_validation)
 
@@ -198,6 +203,9 @@ class PotsdamNumpyFileGenerator(PotsdamFileGenerator):
         _preprocess(TRAIN)
         _preprocess(VALIDATION)
         _preprocess(TEST)
+
+        PotsdamNumpyFileGenerator(
+            datasets_path, [0, 1, 2, 3, 4]).write_channel_stats(proc_data_path)
 
     def get_file_path(self, file_ind):
         ind0, ind1 = file_ind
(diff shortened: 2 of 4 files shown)


lewfish:
One thing we forgot to include with the PR message is the testing instructions.

src/rastervision/common/data/generators.py
```diff
@@ -66,11 +69,22 @@ def __init__(self, active_input_inds, train_ratio, cross_validation):
...
 
-        gen = self.make_split_generator(
-            TRAIN, target_size=(10, 10), batch_size=100, shuffle=True,
-            augment=False, normalize=False, only_xy=False)
-        batch = next(gen)
-        self.normalize_params = get_channel_stats(batch.all_x)
+        """
+        if a specific dataset's normalized parameters have already been
+        calculated, then load that file. Otherwise, calculate with small batch.
+        """
```
lossyrob: The triple quote way of commenting is for docstrings, but for middle of code commends I would use `#`

src/rastervision/common/data/generators.py
```diff
@@ -250,3 +264,14 @@ def transform_img_batch(x):
         split_gen = map(transform_img_batch, img_batch_gen)
 
         return split_gen
+
+    def normalize_constants(self, datasets_path):
```
lossyrob: This name feels a bit off...it took me a bit of reading the code to see what it was doing. Can we come up with a better name? `write_normalization_constants`? Maybe `write_normalization_params` , as that jives more with the `normalization_params` that are loaded from this file. I can see that there's a differentiation between const vs params, but I'd say that if this is simply a serialization of the normalization params, we can name it all `normalization_params`

lewfish: How about `write_channel_stats` to be consistent with the method name?

src/rastervision/common/data/generators.py
```diff
@@ -1,8 +1,9 @@
 import numpy as np
+import codecs, json
+from os.path import exists, join
 
-from rastervision.common.utils import get_channel_stats
 from rastervision.common.settings import TRAIN, VALIDATION, TEST
-
+from rastervision.common.utils import get_channel_stats, save_json
```
lewfish: There's a standard order for grouping imports. First are packages that are built in to Python. Second are third party packages. Third are packages within your project. I've also been grouping our own packages by common, semseg or tagging, just to make it a little easier to read.

src/rastervision/common/data/generators.py
```diff
@@ -66,11 +67,22 @@ def __init__(self, active_input_inds, train_ratio, cross_validation):
...
-        gen = self.make_split_generator(
-            TRAIN, target_size=(10, 10), batch_size=100, shuffle=True,
-            augment=False, normalize=False, only_xy=False)
-        batch = next(gen)
-        self.normalize_params = get_channel_stats(batch.all_x)
+        """
+        if a specific dataset's normalized parameters have already been
+        calculated, then load that file. Otherwise, calculate with small batch.
+        """
+        const_path = join(self.dataset_path, self.name+'_norm_constants.json')
```
lewfish: I'm surprised the linter didn't say to do this, but there should be spaces around the `+`

src/rastervision/common/data/generators.py
```diff
@@ -66,11 +67,22 @@ def __init__(self, active_input_inds, train_ratio, cross_validation):
...
+        if exists(const_path):
+            obj_text = codecs.open(const_path, 'r', encoding='utf-8').read()
```
lewfish: Is there a reason for using the codecs package? If not, I'd prefer loading json according to prior examples in the repo such as https://github.com/azavea/raster-vision/blob/076b5f9a3b381a3c256916644a6f61bcee8c1150/src/rastervision/run.py#L32-L33

src/rastervision/common/data/generators.py
```diff
@@ -66,11 +67,22 @@ def __init__(self, active_input_inds, train_ratio, cross_validation):
...
+            const_list = json.loads(obj_text)
+            const_numpy = np.array(const_list)
+            self.normalize_params = (const_numpy[0], const_numpy[1])
```
lewfish: Along the lines of the previous comment, I would change these to `channel_stats_list`  and `channel_stats`. Since most of the data is this project is in numpy arrays, I would omit the `numpy` suffix.

lewfish: And yes, I do realize that I'm the one who came up with the `normalize_params` name in the first place :)

src/rastervision/common/data/generators.py
```diff
@@ -250,3 +262,14 @@ def transform_img_batch(x):
...
+    def normalize_constants(self, datasets_path):
+        gen = self.make_split_generator(
+            TRAIN, target_size=(10, 10), batch_size=1000, shuffle=True,
+            augment=False, normalize=False, only_xy=False)
+        batch = next(gen)
+        means, stds = get_channel_stats(batch.all_x)
+        const_dict = {'means': means.tolist(), 'stds': stds.tolist()}
+        const_path = join(datasets_path, self.name+'_norm_constants.json')
+        json.dump(const_dict, codecs.open(const_path, 'w', encoding='utf-8'), \
+                  separators=(',', ':'), sort_keys=True, indent=4)
```
lewfish: Do you need the separator argument here?

src/rastervision/common/data/generators.py
```diff
@@ -250,3 +262,14 @@ def transform_img_batch(x):
...
         return split_gen
+
+    def normalize_constants(self, datasets_path):
+        gen = self.make_split_generator(
+            TRAIN, target_size=(10, 10), batch_size=1000, shuffle=True,
+            augment=False, normalize=False, only_xy=False)
+        batch = next(gen)
+        means, stds = get_channel_stats(batch.all_x)
+        const_dict = {'means': means.tolist(), 'stds': stds.tolist()}
+        const_path = join(datasets_path, self.name+'_norm_constants.json')
```
lewfish: Also need space around + here

src/rastervision/common/data/generators.py
```diff
@@ -66,11 +67,22 @@ def __init__(self, active_input_inds, train_ratio, cross_validation):
         if self.cross_validation is not None:
             self.process_cross_validation()
 
-        gen = self.make_split_generator(
-            TRAIN, target_size=(10, 10), batch_size=100, shuffle=True,
-            augment=False, normalize=False, only_xy=False)
-        batch = next(gen)
-        self.normalize_params = get_channel_stats(batch.all_x)
+        """
```
lewfish: Consider putting this stuff in a method called `get_channel_stats` which is distinguished from your other method which actually writes stats to disk.

src/rastervision/semseg/data/potsdam.py
```diff
@@ -74,6 +74,7 @@ class PotsdamImageFileGenerator(PotsdamFileGenerator):
     def __init__(self, datasets_path, active_input_inds,
                  train_ratio=0.8, cross_validation=None):
         self.dataset_path = join(datasets_path, POTSDAM)
+        self.name = "potsdam_image"
```
lewfish: Consider computing this string using the `POTSDAM` and `IMAGE` constants.

lewfish: Nevermind... I see that `POTSDAM` has the `isprs/`  in it.

src/rastervision/semseg/data/test/test_batch_norms.py
```diff
@@ -0,0 +1,43 @@
...
+
+from rastervision.semseg.data.potsdam import (PotsdamNumpyFileGenerator,
+    PotsdamImageFileGenerator)
+from rastervision.common.utils import get_channel_stats
+from rastervision.common.settings import datasets_path
+from rastervision.common.settings import (
+    TRAIN, VALIDATION, TEST
+)
+
+PROCESSED_POTSDAM = 'isprs/processed_potsdam'
```
lewfish: Instead of redefining it, you can just import this constant.

src/rastervision/semseg/data/test/test_batch_norms.py
```diff
@@ -0,0 +1,43 @@
...
+class NormalizationTestCase(unittest.TestCase):
+    def test_numpy_potsdam_batch(self):
+        generator = PotsdamNumpyFileGenerator(datasets_path, [0, 1, 2, 3, 4])
+        gen = generator.make_split_generator(
+            TRAIN, target_size=(10, 10), batch_size=100, shuffle=True,
+            augment=False, normalize=True, only_xy=False)
+        batch = next(gen)
+        means, stds = get_channel_stats(batch.all_x)
+
+        print(means, stds)
```
lewfish: You should remove print statements that were just to help debugging before committing.

src/rastervision/semseg/data/test/test_batch_norms.py
```diff
@@ -0,0 +1,43 @@
...
+        # passes when mean = 0 with an error of +/- 0.3 and stds = 1 +/- 0.25.
+        self.assertTrue((means > -.3).all() and (means < .3).all())
```
lewfish: You could simplify this by saying `(np.absolute(means) < 0.3).all()`. BTW, I think it's more readable when the number is preceded by `0.` but I'm not sure if this is a widely followed convention.

src/rastervision/semseg/data/test/test_batch_norms.py
```diff
@@ -0,0 +1,43 @@
...
+        generator = PotsdamImageFileGenerator(datasets_path, [0, 1, 2, 3, 4])
+        gen = generator.make_split_generator(
+            TRAIN, target_size=(10, 10), batch_size=100, shuffle=True,
+            augment=False, normalize=True, only_xy=False)
+        batch = next(gen)
+        means, stds = get_channel_stats(batch.all_x)
+        print(means, stds)
+        # passes when mean = 0 and stds = 1 with an error of +/- 0.25.
+        self.assertTrue((means > -.25).all() and (means < .25).all())
+        self.assertTrue((stds > .75).all() and (stds < 1.25).all())
```
lewfish: It looks like there's some duplicated code between these two methods. You could create a helper method with the bulk of the functionality that has the class name and the tolerances as parameters, and just call it from these two methods.

lewfish:
There are three things we forgot to do: upload the channel stats files to S3, test it on AWS, and update the README.md file to say that those files need to be put on S3.

lewfish:
The changes you made look good.

lewfish:
I tested this locally on Potsdam and it worked.

lewfish:
Nice job!

lewfish:
After merging a PR it's a good idea to delete the branch. There's a button for doing that in the PR.

PR #42, opened 2017-06-13
WIP: wide-res-net

citerana:


lewfish:
We can keep the az/wide-res-net branch around for now, but this wip PR isn't needed anymore, so I'm closing it without merging.

PR #50, opened 2017-06-21
Add Jupyter and experiments for investigating tiffs and jpegs

citerana:
In order to better understand the difference in performance on the tif versus jpg versions of the Planet Kaggle dataset, this PR adds the ability to use Jupyter notebooks while in the vagrant environment. This also includes several experiments for image performance comparisons and a change to the test-tif path to the new files that now correctly match with the test-jpg files.

You can run Jupyter while in the vagrantbox by using the command `./scripts/jupyter`. Similar to the datasets, this functionality requires that the file path to your local notebooks folder be stored in a global variable under `RASTER_VISION_NOTEBOOK_DIR` in `~/.bash.rc`. The notebooks path in the Docker is at `opt/notebooks`. Note that Jupyter cannot be run concurrently with `./scripts/run --cpu` as both docker contains require the same port.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -31,7 +31,7 @@ Vagrant.configure("2") do |config|
     raster_vision.vm.synced_folder DATA_DIR, "/opt/data"
 
     # If RASTER_VISION_NOTEBOOK_DIR is set, sync it to the VM
-    if File.directory?(File.expand_path(NOTEBOOK_DIR))
+    if NOTEBOOK_DIR and File.directory?(File.expand_path(NOTEBOOK_DIR))
       raster_vision.vm.synced_folder NOTEBOOK_DIR, "/opt/notebooks"
     end
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -11,8 +11,13 @@ SCRIPT
 
 DATA_DIR = ENV["RASTER_VISION_DATA_DIR"]
 if DATA_DIR.nil?
-  puts "You must set the environment variable: RASTER_VISION_DATA_DIR"
+  puts "ERROR: You must set the environment variable: RASTER_VISION_DATA_DIR"
   exit 1
+end
+
+NOTEBOOK_DIR = ENV["RASTER_VISION_NOTEBOOK_DIR"]
+if NOTEBOOK_DIR.nil?
+  puts "WARN: To use juptyer, You must set the environment variable: RASTER_VISION_NOTEBOOK_DIR"
 end
 
 S3_BUCKET = ENV.fetch("RASTER_VISION_BUCKET", "raster-vision")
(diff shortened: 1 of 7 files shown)


scripts/jupyter
```diff
@@ -0,0 +1,33 @@
...
+}
+
+if [ "${BASH_SOURCE[0]}" = "${0}" ]
+then
+    docker run --rm -it \
+           -v "$SRC":/opt/src \
+           -v /opt/data:/opt/data \
+           -v /opt/notebooks:/opt/notebooks \
+           -p 8888:8888 \
+           raster-vision-base \
```
lewfish: I think this should be changed to `raster-vision-cpu` so that we can access Tensorflow from inside the notebook. (The base image doesn't have TF.)

src/Dockerfile-base
```diff
@@ -35,6 +35,8 @@ RUN conda install -y python=${python_version} scikit-learn six h5py \
         matplotlib pillow rasterio jupyter scikit-image && \
     conda clean -yt
 
+RUN pip install spectral
```
lewfish: Consider removing this line and just adding `spectral` to the next line which runs `pip install`.

src/experiments/tagging/6_19_17/IRRGtiff.json
```diff
@@ -0,0 +1,18 @@
+{
```
lewfish: I would say that the first two sets of experiment files (from 6//19) had some bugs in them, so they shouldn't be committed.

src/Dockerfile-base
lewfish: When you add a dependency, it's a good idea to say in the commit message why that dependency was added.

src/rastervision/tagging/data/planet_kaggle.py
```diff
@@ -282,10 +282,9 @@ def make_batch(self, img_batch, file_inds):
 class PlanetKaggleTiffFileGenerator(PlanetKaggleFileGenerator):
     def __init__(self, datasets_path, options):
         self.dev_dir = 'train-tif-v2'
-        self.test_dir = 'test-tif-v2'
-        self.file_names = [
-            'train-tif-v2.zip', 'test-tif-v2.zip', 'train_v2.csv.zip',
-            'planet_kaggle_tiff_channel_stats.json']
+        self.test_dir = 'test-tif-v3'
+        self.zip_file_names = [
```
lewfish: When resolving merge conflicts, you mistakenly changed `file_names` to `zip_file_names` and also omitted the channel stats file.

lewfish:
Testing it now..

Vagrantfile
```diff
@@ -25,6 +30,11 @@ Vagrant.configure("2") do |config|
     raster_vision.vm.synced_folder "~/.aws", "/home/vagrant/.aws"
     raster_vision.vm.synced_folder DATA_DIR, "/opt/data"
 
+    # If RASTER_VISION_NOTEBOOK_DIR is set, sync it to the VM
+    if File.directory?(File.expand_path(NOTEBOOK_DIR))
```
lewfish: If notebook dir isn't set, which is only considered a warning, then this line will fail since it will try to call `expand_path` on nil. This line should be changed to 
```
    if NOTEBOOK_DIR and File.directory?(File.expand_path(NOTEBOOK_DIR))
``` 
to avoid this.

lewfish:
Everything else works. After you make that change, I'll give it a final test.

lewfish:
BTW, it's easier to review changes if you put them into a separate fixup commit(s), which you can then rebase before merging.

lewfish:
👍  Remember to squash the fixup commit before merging.

PR #54, opened 2017-06-26
Update to scoring file

citerana:
This PR updates the scores.json file created during the validation_eval task. It now displays the f2 score weighted across labels, the subscores grouped according to tag and all scores are displayed to 5 decimal places of precision.

### DIFF
--- a/src/rastervision/tagging/tasks/validation_eval.py
+++ b/src/rastervision/tagging/tasks/validation_eval.py
@@ -21,11 +21,12 @@ VALIDATION_EVAL = 'validation_eval'
 class Scores():
     """A set of scores for the performance of a model on a dataset."""
     def __init__(self, y_true, y_pred, dataset):
-        self.f2 = fbeta_score(y_true, y_pred, beta=2, average='samples')
-        self.f2_weighted = fbeta_score(y_true, y_pred, beta=2,
-                                       average='weighted')
+        self.f2_samples = fbeta_score(y_true, y_pred, beta=2, average='samples')
+        self.f2_labels = fbeta_score(y_true, y_pred, beta=2,
+                                       average='macro')
         self.atmos_scores, self.common_scores, self.rare_scores = {},{},{}
         f2_subscores = fbeta_score(y_true, y_pred, beta=2, average=None)
+        # TODO: update all_tags usage with active_tags PR
         for tag in dataset.atmos_tags:
             self.atmos_scores[tag] = f2_subscores[dataset.all_tags.index(tag)]
         for tag in dataset.common_tags:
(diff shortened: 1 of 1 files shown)


src/rastervision/tagging/tasks/validation_eval.py
```diff
@@ -18,13 +20,21 @@
 
 class Scores():
     """A set of scores for the performance of a model on a dataset."""
-    def __init__(self, y_true, y_pred, all_tags):
+    def __init__(self, y_true, y_pred, dataset):
         self.f2 = fbeta_score(y_true, y_pred, beta=2, average='samples')
+        self.f2_weighted = fbeta_score(y_true, y_pred, beta=2,
+                                       average='weighted')
```
lewfish: I think we want to use the `macro` option for `average`. This will not account for unbalanced labels which is what we want.

lewfish: Also, it would be more user-friendly if we renamed `f2` to `f2_samples` and `f2_weighted` to `f2_labels`

src/rastervision/tagging/tasks/validation_eval.py
```diff
@@ -18,13 +20,21 @@
...
+        self.atmos_scores, self.common_scores, self.rare_scores = {},{},{}
         f2_subscores = fbeta_score(y_true, y_pred, beta=2, average=None)
-        self.f2_subscores = dict(zip(all_tags, f2_subscores))
+        for tag in dataset.atmos_tags:
+            self.atmos_scores[tag] = f2_subscores[dataset.all_tags.index(tag)]
```
lewfish: This section of code is going to have to change after combining with the PR on active_tags, since the `all_tags.index` will no longer be a valid way to get indices. Not sure who will merge their PR first, but it would be a good idea to insert a TODO to update this.

src/rastervision/tagging/tasks/validation_eval.py
```diff
@@ -18,13 +20,21 @@
...
+        for tag in dataset.common_tags:
+            self.common_scores[tag] = f2_subscores[dataset.all_tags.index(tag)]
+        for tag in dataset.rare_tags:
+            self.rare_scores[tag] = f2_subscores[dataset.all_tags.index(tag)]
 
     def to_json(self):
-        return json.dumps(self.__dict__, sort_keys=True, indent=4)
+        return json.dumps(self.__dict__, sort_keys=False, indent=4)
```
lewfish: I think you had set this to False back when you were grouping things by the order within the f2_subscores, but we should go back to the old way now that you're grouping them using separate dictionaries.

PR #55, opened 2017-06-27
Add DenseNet models and experiments

citerana:
This PR adds the densenet121 and densenet169 models to the project. Both models support pretraining with imagenet weights. Additionally, this includes a series of experiments that were used to investigate the relative performances of the different architectures with variable learning rate schedules.

The link to the imagenet weights can be found at the top level comment of each model. These models use a tensorflow backend so choose the weights accordingly. They must be accessible by the Docker container via the file path `/opt/data/weights/`.

### DIFF
--- a/src/rastervision/common/models/custom_layers.py
+++ b/src/rastervision/common/models/custom_layers.py
@@ -1,0 +1,75 @@
+# flake8: noqa
+# -*- coding: utf-8 -*-
+"""
+Custom Scale Layer for Keras
+This code is based on
+https://github.com/flyyufelix/DenseNet-Keras
+"""
+
+from keras import initializers
+from keras.engine import Layer, InputSpec
+
+import keras.backend as K
+
+class Scale(Layer):
+    '''Custom Layer for DenseNet used for BatchNormalization.
+
+    Learns a set of weights and biases used for scaling the input data.
+    the output consists simply in an element-wise multiplication of the input
+    and a sum of a set of constants:
+
+        out = in * gamma + beta,
+
+    where 'gamma' and 'beta' are the weights and biases larned.
+
+    # Arguments
+        axis: integer, axis along which to normalize in mode 0. For instance,
+            if your input tensor has shape (samples, channels, rows, cols),
+            set axis to 1 to normalize per feature map (channels axis).
+        momentum: momentum in the computation of the
+            exponential average of the mean and standard deviation
+            of the data, for feature-wise normalization.
+        weights: Initialization weights.
+            List of 2 Numpy arrays, with shapes:
+            `[(input_shape,), (input_shape,)]`
+        beta_init: name of initialization function for shift parameter
+            (see [initializers](../initializers.md)), or alternatively,
+            Theano/TensorFlow function to use for weights initialization.
+            This parameter is only relevant if you don't pass a `weights` argument.
+        gamma_init: name of initialization function for scale parameter (see
+            [initializers](../initializers.md)), or alternatively,
+            Theano/TensorFlow function to use for weights initialization.
+            This parameter is only relevant if you don't pass a `weights` argument.
+    '''
+    def __init__(self, weights=None, axis=-1, momentum = 0.9, beta_init='zero', gamma_init='one', **kwargs):
+        self.momentum = momentum
+        self.axis = axis
+        self.beta_init = initializers.get(beta_init)
+        self.gamma_init = initializers.get(gamma_init)
+        self.initial_weights = weights
+        super(Scale, self).__init__(**kwargs)
+
+    def build(self, input_shape):
+        self.input_spec = [InputSpec(shape=input_shape)]
+        shape = (int(input_shape[self.axis]),)
+
+        self.gamma = K.variable(self.gamma_init(shape), name='{}_gamma'.format(self.name))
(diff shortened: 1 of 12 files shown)


lewfish:
> They must be accessible by the docker via the file path /opt/data/weights/

FYI, people usually refer to a container as a "container" or a "Docker container", but not as a "docker".

lewfish:
> You may optionally connect an AWS bucket and download the weights if they do not already exist locally. In order to do so, the docker must be given valid AWS credentials.

I think we should just copy the file over via the console, or via the command line in the Vagrant box. Giving credentials to the container is more trouble than it's worth and could cause some unintended consequences given the way things are written right now.

src/rastervision/common/models/custom_layers.py
```diff
@@ -0,0 +1,67 @@
+from keras import initializers
```
lewfish: You should add comments at the top of these three files noting where they were modified from as well as the license of the repo they come from.

lewfish: I see you did it for the other two files ... just might want to move it to the top since the stuff we copied spans more than the function that is commented.

src/rastervision/common/models/densenet121.py
```diff
@@ -0,0 +1,195 @@
...
+
+    x_fc = GlobalAveragePooling2D(name='pool'+str(final_stage))(x)
+    x_fc = Dense(1000, name='fc6')(x_fc)
+    x_fc = Activation('softmax', name='prob')(x_fc)
+
+    model = Model(img_input, x_fc, name='densenet')
+
+    # load weights
+    if weights == 'imagenet':
+        weights_path = '/opt/data/weights/densenet121_weights_tf.h5'
```
lewfish: Instead of hardcoding this path, you should do something like `join(weights_path, 'densenet121_weights_tf.h5')`.

src/rastervision/common/models/densenet169.py
```diff
@@ -0,0 +1,195 @@
...
+
+    x_fc = GlobalAveragePooling2D(name='pool'+str(final_stage))(x)
+    x_fc = Dense(1000, name='fc6')(x_fc)
+    x_fc = Activation('softmax', name='prob')(x_fc)
+
+    model = Model(img_input, x_fc, name='densenet')
+
+    # load weights
+    if weights == 'imagenet':
+        weights_path = '/opt/data/weights/densenet169_weights_tf.h5'
```
lewfish: Same thing here.

src/rastervision/tagging/models/factory.py
```diff
@@ -1,7 +1,11 @@
 from rastervision.common.models.factory import ModelFactory
 from rastervision.common.models.resnet50 import ResNet50
+from rastervision.common.models.densenet121 import DenseNet121
+from rastervision.common.models.densenet169 import DenseNet169
 
 BASELINE_RESNET = 'baseline_resnet'
+DENSENET_121 = "densenet121"
+DENSENET_169 = "densenet169"
```
lewfish: Double quotes should be single quotes.

src/rastervision/common/models/densenet121.py
```diff
@@ -1,4 +1,13 @@
-# -*- coding: utf-8 -*-
+'''
+-*- coding: utf-8 -*-
```
lewfish: What does this line mean?

src/rastervision/common/models/densenet169.py
```diff
@@ -85,10 +90,10 @@ def DenseNet169(nb_dense_block=4, growth_rate=32, nb_filter=64, reduction=0.5,
 
     # load weights
     if weights == 'imagenet':
-        weights_path = '/opt/data/weights/densenet169_weights_tf.h5'
-        if not exists(weights_path):
+        path = join(weights_path, 'densenet169_weights_tf.h5')
+        if not exists(path):
             download_weights("densenet169_weights_tf.h5")
```
lewfish: Some more double quotes. Maybe you should do a search and replace on these files.

lewfish:
After running `./scripts/test` a whole bunch of lint errors show up. I think we can just suppress the errors for the files you copied and pasted by adding this line `# flake8: noqa` to the top of the file which is done in https://github.com/azavea/raster-vision/blob/develop/src/rastervision/common/models/resnet50.py. But you should fix the errors for the other files.

src/rastervision/common/models/densenet121.py
citerana: As a comment, it requires that the file be read by a certain python codec. I don't think it does any harm as it's only in the copied files.

src/rastervision/common/models/densenet121.py
```diff
@@ -0,0 +1,200 @@
...
+
+    model = Model(img_input, x_fc, name='densenet')
+
+    # load weights
+    if weights == 'imagenet':
+        path = join(weights_path, 'densenet121_weights_tf.h5')
+        if not exists(path):
+            download_weights("densenet121_weights_tf.h5")
+        model.load_weights(path, by_name=True)
+
```
citerana: fixed - flake8 was crashing on some of the copied files so it had been auto-disabled...

lewfish:
It's easier to review if there is a separate commit added for each round of review.

PR #59, opened 2017-06-28
Drop unwanted tiff files; experiments with tiff files

lossyrob:
This PR does several things:

- Adds `eprint` utility function, which allows you to print things to sterr and therefore have the messages show up in the cloudwatch logs
- upgrades daul_fcn_resnet to latest TF version
- creates a dual_resnet model, PlanetKaggleDualFileGenerator and DualDataset for training on both tiffs and jpgs
- Allows the data generator to specify a drop_file_inds path, which selectively removes some of the file_inds
- Allows for data generators to work with more than one file extension
- Allows for data generators to work with more than one training and test directory
- Don't require that data generators have an ndvi band if they have more than 3 bands
- Adds a slew of experiments trying to get some performance out of the ill-fated TIFF files.


src/rastervision/tagging/tasks/validation_eval.py
```diff
@@ -65,9 +65,10 @@ def plot_prediction(generator, all_x, y_true, y_pred,
         imgs.append(ir_im)
         titles.append('IR')
 
-        ndvi_im = all_x[:, :, dataset.ndvi_ind]
-        imgs.append(ndvi_im)
-        titles.append('NDVI')
+        if dataset.ndvi_ind:
```
lewfish: This is problematic since ndvi_ind could be zero, in which case the condition would be false. Instead, use `dataset.ndvi_ind is not None`

src/rastervision/tagging/models/dual_resnet.py
```diff
@@ -0,0 +1,98 @@
...
+
+def make_dual_resnet(input_shape, dual_active_input_inds,
+                     weights='imagenet', freeze_base=False,
+                     include_top=True, pooling=None,
+                     merge_stage=None,
+                     classes=1000, activation='softmax'):
+    nb_rows, nb_cols, nb_channels = input_shape
+    input_tensor = Input(shape=input_shape)
+
+    if not merge_stage:
```
lewfish: Same issue with merge_stage possibly being zero.

src/rastervision/tagging/models/dual_resnet.py
```diff
@@ -0,0 +1,98 @@
...
+            layer.trainable = False
+
+    base_model2 = ResNet50(
+        include_top=False, weights=weights, input_tensor=input_tensor2)
+    if freeze_base:
+        for layer in base_model2.layers:
+            layer.trainable = False
+
+    for layer in base_model2.layers:
+        layer.name += '_2'
```
lewfish: Consider using same order and spacing for base_model2 code block as for base_model1

src/rastervision/tagging/models/dual_resnet.py
```diff
@@ -0,0 +1,98 @@
...
+        x_1 = base_model1.get_layer('act4f_1').output
+        x_2 = base_model2.get_layer('act4f_2').output
+
+        x = merge([x_1, x_2], mode='concat')
+
+        x = conv_block(x, 3, [512, 512, 2048], stage=5, block='a')
+        x = identity_block(x, 3, [512, 512, 2048], stage=5, block='b')
+        x = identity_block(x, 3, [512, 512, 2048], stage=5, block='c')
+    else:
+        x_1 = base_model1.get_layer('act5c_1').output
```
lewfish: Consider only doing this if `merge_stage == 5` and then throw an error for any other value.

src/rastervision/tagging/models/dual_resnet.py
```diff
@@ -0,0 +1,98 @@
...
+        x = Flatten()(x)
+        x = Dense(classes, activation=activation, name='dense')(x)
+    else:
+        if pooling == 'avg':
+            x = GlobalAveragePooling2D()(x)
+        elif pooling == 'max':
+            x = GlobalMaxPooling2D()(x)
+
+
+    model = Model(input_tensor, output=x)
```
lewfish: Looks like there are some lint errors. Run `./scripts/test` to see them.

src/rastervision/tagging/data/planet_kaggle.py
```diff
@@ -218,11 +226,28 @@ def __init__(self, datasets_path, options):
...
-        self.dev_file_inds = self.generate_file_inds(self.dev_path)
-        self.test_file_inds = self.generate_file_inds(self.test_path)
+        self.drop_file_inds = []
+        if self.drop_file_inds_file:
+            with open(join(self.dataset_path, self.drop_file_inds_file)) as f:
+                self.drop_file_inds = f.read().split('\n')
+
+        # dev_dir and test_ dir can contain multiple directories
+        # Assume file_inds are the same in every dev_dir and test_dir, respectively
+        if isinstance(self.dev_dir, list):
```
lewfish: There are several places where we test if something is a list and do one thing, and then another if it isn't. I think it would be simpler to make it so that `dev_dir` is always a list, and then we don't need two code paths. This means that wherever `dev_dir` was a singleton, we just replace with a list of length one.

src/rastervision/tagging/data/planet_kaggle.py
```diff
@@ -287,10 +321,22 @@ def plot_sample(self, file_path, x, y, file_ind):
...
-            self.dataset_path, data_dir, '{}.{}'.format(
-                file_ind, self.file_extension))
+                                  else self.dev_path
+        def _get_file_path(d):
+            if isinstance(self.file_extension, list):
+                for fe in self.file_extension:
+                    n = '{}.{}'.format(file_ind, fe)
+                    p = join(self.dataset_path, d, n)
+                    if exists(p):
+                        return p
```
lewfish: More verbose variable names would help readability here.

lewfish:
Should we also include the script that dropped the tiffs?

src/rastervision/common/options.py
```diff
@@ -58,3 +58,7 @@ def __init__(self, options):
                 raise ValueError(
                     '{} are not valid augment_methods'.format(
                         str(invalid_augment_methods)))
+
+
+        self.dual_active_input_inds = options.get('dual_active_input_inds')
```
lossyrob: We had this setting before for ISPRS, and it got dropped. We should keep this change for backward compatibility

src/rastervision/common/utils.py
```diff
@@ -15,6 +15,8 @@
     s3_results_path, results_path, s3_datasets_path, datasets_path,
     s3_weights_path, weights_path, s3_bucket)
 
+def eprint(*args, **kwargs):
```
lossyrob: eprint is useful in that it prints out to cloudwatch logs; perhaps we should leave this until we have a better logging mechanism?

src/rastervision/common/utils.py
```diff
@@ -198,6 +200,9 @@ def get_file(file_name):
 
         for file_name in file_names:
             get_file(file_name)
+    eprint("Dataset directory files in %s:" % dataset_path)
+    for x in os.listdir(dataset_path):
+        eprint(" - " + x)
```
lossyrob: This was useful in debugging when files were not being pulled down correctly.

src/rastervision/semseg/models/dual_fcn_resnet.py
```diff
@@ -23,13 +23,11 @@ def make_dual_fcn_resnet(input_shape, dual_active_input_inds,
...
     def get_input_tensor(it, model_ind):
-        # TODO calls to split and concat will need to be updated after
-        # upgrading TF
-        channel_tensors = tf.split(3, nb_channels, it)
+        channel_tensors = tf.split(it, nb_channels, 3)
         input_tensors = []
         for ind in dual_active_input_inds[model_ind]:
             input_tensors.append(channel_tensors[ind])
-        input_tensor = tf.concat(3, input_tensors)
+        input_tensor = tf.concat(input_tensors, 3)
```
lossyrob: Updates to dual_fcn_resnet should stay, removes TODOs

src/rastervision/tagging/data/factory.py
```diff
@@ -23,6 +23,8 @@ def get_class(self, dataset_name, generator_name):
                 return PlanetKaggleTiffFileGenerator
             elif generator_name == JPG:
                 return PlanetKaggleJpgFileGenerator
+            elif generator_name == DUAL:
+                return PlanetKaggleDualFileGenerator
```
lossyrob: we can drop all the dual file generator/DUAL code

src/rastervision/tagging/data/planet_kaggle.py
lossyrob: That makes sense. However since we're dropping the TIFF stuff, it might be best to just leave it as a non-list.

src/rastervision/tagging/data/planet_kaggle.py
lossyrob: Change can be dropped

src/rastervision/tagging/data/planet_kaggle.py
```diff
@@ -316,10 +362,11 @@ def __init__(self, datasets_path, options):
         self.test_dir = 'test-tif-v3'
         self.file_names = [
             'train-tif-v2.zip', 'test-tif-v3.zip', 'train_v2.csv.zip',
-            'planet_kaggle_tiff_channel_stats.json']
+            'planet_kaggle_tiff_channel_stats.json', 'unaligned_tifs.csv']
```
lossyrob: unaligned tifs didn't work out, related changes can be dropped

src/rastervision/tagging/data/planet_kaggle.py
```diff
@@ -388,3 +438,43 @@ def __init__(self):
         options = Options()
         PlanetKaggleJpgFileGenerator(
             datasets_path, options).write_channel_stats(proc_data_path)
+
+class PlanetKaggleDualFileGenerator(PlanetKaggleFileGenerator):
```
lossyrob: drop change

src/rastervision/tagging/models/dual_resnet.py
```diff
@@ -0,0 +1,98 @@
+"""
+Late Fusion of 2 stream ResNet
+"""
+from keras.models import Model
```
lossyrob: Can drop this entire file.

src/rastervision/tagging/models/factory.py
```diff
@@ -2,6 +2,7 @@
 from rastervision.common.models.resnet50 import ResNet50
 from rastervision.common.models.densenet121 import DenseNet121
 from rastervision.common.models.densenet169 import DenseNet169
+from rastervision.tagging.models.dual_resnet import (make_dual_resnet, DUAL_RESNET)
```
lossyrob: revert changes to this file.

src/rastervision/tagging/tasks/validation_eval.py
lossyrob: can revert these changes.

lewfish:
The only changes we want to keep are in `common/utils.py` and `semseg/models/dual_fcn_resnet.py`, so we have extracted them out into a new PR https://github.com/azavea/raster-vision/pull/91

PR #61, opened 2017-06-29
Perform safe augmentations to test images during prediction

citerana:
This PR add the ability to perform a series of safe augmentations to each test image we want to predict label probabilities for. The augmentations are 90 degree rotations, vertical and horizontal flipping. We submit each augmented image to be predicted on by a trained model, take the resulting predictions of each augmented image and averages the probabilities for a more accurate final prediction. This is done during `compute_probs` only during the testing split. For probability calculations during training and validation, we still submit images for prediction only once.

Test-time augmentation is performed by default, however you may choose to perform only simple single predictions for an experiment by adding the line `"test_augmentation": false,` in the experiment file.

### DIFF
--- a/src/rastervision/tagging/tasks/predict.py
+++ b/src/rastervision/tagging/tasks/predict.py
@@ -9,6 +9,7 @@ from rastervision.common.options import AGG_CONCAT, AGG_ENSEMBLE
 from rastervision.tagging.data.planet_kaggle import TagStore
 from rastervision.tagging.tasks.utils import compute_prediction
 from rastervision.tagging.tasks.train_thresholds import load_thresholds
+from rastervision.common.settings import TEST
 
 TRAIN_PROBS = 'train_probs'
 VALIDATION_PROBS = 'validation_probs'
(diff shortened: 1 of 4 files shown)


src/rastervision/tagging/tasks/predict.py
```diff
@@ -87,7 +88,27 @@ def compute_probs(run_path, model, options, generator, split):
...
+        # probs and taking average, if calculating test probabilities
+        if split == TEST:
+            for img_ind in range(batch.x.shape[0]):
+                img = batch.x[img_ind, :, :, :]
+                rot_batch = []
+                for rotation in range(4):
+                    rot_img = np.rot90(img, rotation)
+                    rot_img = np.expand_dims(rot_img, axis=0)
+                    rot_batch.append(rot_img)
+                rot_batch.append(np.flipud(rot_img))
```
lewfish: It turns out that `rot_img` is the same as `img` after rotating it 4 times, but I think it would be clearer to call `flipud` on `img`. Same for the next line.

src/rastervision/tagging/tasks/predict.py
```diff
@@ -87,7 +88,27 @@ def compute_probs(run_path, model, options, generator, split):
...
+                rot_batch.append(np.fliplr(rot_img))
+                rot_batch = np.concatenate(rot_batch, axis=0)
+
+                rot_probs = np.mean(model.predict(rot_batch), axis=0)
+                rot_probs = np.expand_dims(rot_probs, axis=0)
```
lewfish: If you do `rot_probs = np.mean(model.predict(rot_batch), axis=0, keepdims=True)`, then you can get rid of this line.

src/rastervision/tagging/tasks/predict.py
```diff
@@ -87,7 +88,27 @@ def compute_probs(run_path, model, options, generator, split):
         normalize=True, only_xy=False)
 
     for batch_ind, batch in enumerate(split_gen):
-        y_probs.append(model.predict(batch.x))
+        # Performing safe augmentations on images one by one, predicting
+        # probs and taking average, if calculating test probabilities
+        if split == TEST:
+            for img_ind in range(batch.x.shape[0]):
+                img = batch.x[img_ind, :, :, :]
+                rot_batch = []
```
lewfish: Instead of `rot_batch` and `rot_probs`, `aug_batch` and `aug_probs` would be more accurate names.

src/experiments/tagging/6_27_17/testrotation_0.json
```diff
@@ -0,0 +1,19 @@
+{
+    "batch_size": 8,
+    "git_commit": "95f4e9",
```
lewfish: I'm not sure what the intention is behind these experiment files, especially why you need four of them to test whether this works. I think you should explain this in the commit message.

lewfish:
Remember to squash commits before merging. Also consider adding a note to the PR description about the results we got with and without the test-time augmentation.

PR #66, opened 2017-07-07
Added option to optionally compute channel statistics

lossyrob:
If you want to use the generators for tasks that do not need to normalize or un-normalize the images, the requirement to compute channel statistics is burdensome (having to download the correct precomputed channel statistics or the imagery). This PR gives an option that allows for the channel statics to not be computed if the tasks do not need them.

### DIFF
--- a/src/rastervision/common/data/generators.py
+++ b/src/rastervision/common/data/generators.py
@@ -81,18 +81,23 @@ class FileGenerator(Generator):
 
         self.train_probs = self.compute_train_probs()
 
-        # If a dataset's normalized parameters have already been
-        # calculated, load its json file. Otherwise, calculate parameters
-        # with a small batch.
-        channel_stats_path = join(self.dataset_path,
-                                  self.name + '_channel_stats.json')
-        if exists(channel_stats_path):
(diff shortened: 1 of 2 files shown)


lewfish:
I can see that there are times when we don't actually need the images to be normalized, or to even iterate over images, but why is it a burden to download the channel_stats file if that happens automatically anyway and is a tiny file?

lossyrob:
It's unclear what the codepath is, but when I don't set this option, my current notebook tries to calculate from the images.

lewfish:
> I think this speaks a bit to that the data generator holds a lot of responsibility, and if it were split out into a file index generator and something that actually loaded and normalized the images, I could use the former class only and avoid any unnecessary downloads.

True, it could use refactoring. So, since it happens automatically do you still need the functionality in this PR?

lossyrob:
The problem was with setting the proper environment variables so that the Jupyter notebook can use the right S3 path; also when downloads fail it continues as if it had succeeded. 

These parts of the code need general refactoring, like you mentioned; this would be a band-aide where more work is needed, so I'll close this out.

PR #69, opened 2017-07-08
Add XGBoost to jupyter and CPU containers

lossyrob:
Also adds pandas and opencv to the CPU container


citerana:
Oh, I just saw but Lewis did mention that this "doesn't allow you to specify the number of CPUs needed. The default is 1 set here https://github.com/azavea/raster-vision/blob/develop/deployment/batch/job_def_cpu.json#L10-11. You'll want to modify `submit_jobs` so that it overrides the number of vCPUS here https://github.com/azavea/raster-vision/blob/develop/scripts/submit_jobs#L111."

PR #70, opened 2017-07-10
Add ability to use YellowFin optimizer

citerana:
This PR adds a Keras-wrapped TensorFlow optimizer called YellowFin. This optimizer automatically adjusts several hyperparameters while the model trains to converge towards some optimal confirguation. YellowFin takes an initial learning rate, but has optionally tuned momentum and learning_factor parameters.

In our experiments with the Planet Kaggle dataset, we found that the recommended `init_lr=1.0` and `momentum=0.0` from the paper performed poorly. It is unclear why YF behaved in such a way, but take the recommendations with a grain of salt and properly tune this optimizer to your models.

### DIFF
--- a/src/rastervision/common/optimizers/yellowfin.py
+++ b/src/rastervision/common/optimizers/yellowfin.py
@@ -1,0 +1,278 @@
+# flake8: noqa
+"""
+YellowFin optimizer for TensorFlow
+Optimizer is based on:
+https://github.com/JianGoForIt/YellowFin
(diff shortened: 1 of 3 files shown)


lewfish:
If you fix merge conflicts, move `yellowfin.py` to `common/optimizers/yellowfin.py`, and add a comment in the file saying that we weren't able to get it to work and that it's "experimental" this should be good to go.

PR #73, opened 2017-07-11
Expand learning rate manipulation options

citerana:
This PR adds several new ways to modify the learning rate throughout the training process.

One of these methods is a cyclic learning rate. For usage, you must specify a subset of four options for `cyclic_lr`, which are `base_lr`, `max_lr`, `step_size` and `cycle_mode`.  These parameters do not have defaults and must all be specified in order to use this functionality. More details on this implementation of cyclic learning rates can be found [here](https://github.com/bckenstler/CLR).

Usage example in an experiment file:

```
"cyclic_lr": { 
      "base_lr": 1e-6, 
      "max_lr": 1e-4, 
      "step_size": 2000, 
      "cycle_mode": "triangular" 
}, 
```

Also, this PR introduces two options for decaying the learning rate as the model trains. You may now give a decay value for optimizers with this parameter using the option `decay`. If you would prefer that the learning rate decay by some factor after each epoch, give the factor instead as `epoch_decay`.

Note that these decay options are mutually exclusive and a maximum of one decay option may be set per experiment.

### DIFF
--- a/src/rastervision/common/tasks/cyclic_lr.py
+++ b/src/rastervision/common/tasks/cyclic_lr.py
@@ -1,0 +1,136 @@
+# flake8: noqa
+"""
+Cyclic Learning Rates for Keras
+Callback Schema is based on
+https://github.com/bckenstler/CLR
+"""
+from keras.callbacks import *
+
+class CyclicLR(Callback):
+    """This callback implements a cyclical learning rate policy (CLR).
+    The method cycles the learning rate between two boundaries with
+    some constant frequency, as detailed in this paper (https://arxiv.org/abs/1506.01186).
+    The amplitute of the cycle can be scaled on a per-iteration or
+    per-cycle basis.
+    This class has three built-in policies, as put forth in the paper.
+    "triangular":
+        A basic triangular cycle w/ no amplitude scaling.
+    "triangular2":
+        A basic triangular cycle that scales initial amplitude by half each cycle.
+    "exp_range":
+        A cycle that scales initial amplitude by gamma**(cycle iterations) at each
+        cycle iteration.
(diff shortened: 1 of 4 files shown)


src/rastervision/common/tasks/cyclic_lr.py
```diff
@@ -0,0 +1,131 @@
+# flake8: noqa
```
lewfish: We need to put a link to the original code at the top of the file.

src/rastervision/common/tasks/train_model.py
```diff
@@ -81,6 +81,13 @@ def get_lr(epoch):
             callback = LearningRateScheduler(get_lr)
             callbacks.append(callback)
 
+        if self.options.epoch_decay:
+            def get_lr(epoch):
+                decay_factor = 1 / (1.0 + self.options.epoch_decay * epoch)
+                return self.options.init_lr * decay_factor
+            callback = LearningRateScheduler(get_lr)
+            callbacks.append(callback)
```
lewfish: It seems like `lr_schedule`, `cyclic_lr`, `decay` and `epoch_decay` should be mutually exclusive options. Otherwise they will conflict with each other. You should raise an error in options if this happens.

src/rastervision/common/tasks/train_model.py
```diff
@@ -81,6 +81,13 @@ def get_lr(epoch):
             callback = LearningRateScheduler(get_lr)
             callbacks.append(callback)
 
+        if self.options.epoch_decay:
```
lewfish: Instead of `decay` and `epoch_decay`, I think `lr_step_decay` and `lr_epoch_decay` would be clearer.

PR #77, opened 2017-07-13
Generate separate files for each augmention

citerana:
In the interest of generating more diversity and being able to better manipulate augmented predictions, we wanted to have the probabilities and predictions obtained from augmented images accessible at all times. This PR creates a probability and prediction file for each separate image augmentation during test time. There is no longer a single `test_preds.csv` prediction file, but six separate prediction files that may be combined or used separately in submission.

Note that the numbers of filenames do consistently correspond with an image augmentation. Files beginning with 0-3 are the results from rotations and files beginning with 4 and 5 are flipped images.

### DIFF
--- a/src/rastervision/tagging/tasks/predict.py
+++ b/src/rastervision/tagging/tasks/predict.py
@@ -18,6 +18,8 @@ TRAIN_PREDICT = 'train_predict'
 VALIDATION_PREDICT = 'validation_predict'
 TEST_PREDICT = 'test_predict'
 
+NUM_AUG = 6
+
 
 def get_probs_fn(split):
     return '{}_probs.npy'.format(split)
@@ -25,6 +27,9 @@ def get_probs_fn(split):
 
 def get_preds_fn(split):
     return '{}_preds.csv'.format(split)
+
+def get_aug_probs_fn(split, aug_ind):
+    return str(aug_ind) + '_' + get_probs_fn(split)
 
 
 def compute_concat_probs(run_path, options, generator, split):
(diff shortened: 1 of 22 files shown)


src/rastervision/tagging/tasks/predict.py
```diff
@@ -101,40 +101,82 @@ def compute_probs(run_path, model, options, generator, split):
...
-    if options.nb_eval_samples is not None:
-        y_probs = y_probs[0:options.nb_eval_samples, :]
+        y_probs = np.concatenate(y_probs, axis=0)
 
-    np.save(probs_path, y_probs)
+        if options.nb_eval_samples is not None:
+            y_probs = y_probs[0:options.nb_eval_samples, :]
 
+    # Saving the probability numpy arrays
+    if split == TEST and options.test_augmentation:
```
lewfish: Instead of repeating this if/else block you can just move the code up into the existing if/else block

src/rastervision/tagging/tasks/predict.py
```diff
@@ -101,40 +101,82 @@ def compute_probs(run_path, model, options, generator, split):
...
+        for aug_ind in range(NUM_AUG):
+            probs_path = join(run_path, str(aug_ind) + '_' +
```
lewfish: Consider moving the code that generates the prob_path into a function since it is used in more than one place in this file.

src/rastervision/tagging/tasks/predict.py
```diff
@@ -18,6 +18,7 @@
 VALIDATION_PREDICT = 'validation_predict'
 TEST_PREDICT = 'test_predict'
 
+NUM_AUG = 6
 
 def get_probs_fn(split):
```
lewfish: Need another new line after NUM_AUG for lint

src/rastervision/tagging/tasks/predict.py
```diff
@@ -101,40 +101,82 @@ def compute_probs(run_path, model, options, generator, split):
...
+
+                for aug_ind in range(NUM_AUG):
+                    y_probs[aug_ind].append(aug_probs[aug_ind])
+
+            if (options.nb_eval_samples is not None and
+                    batch_ind * options.batch_size >= options.nb_eval_samples):
+                break
+
+        for aug_ind in range(NUM_AUG):
+            y_probs[aug_ind] = np.concatenate([y_probs[aug_ind]])
```
lewfish: I don't think you want to enclose `y_probs[aug_ind]` in `[]`. `np.concatenate` expects a list, and `y_probs[aug_ind]` already is a list.

citerana: Intuitively I would agree with you, but I actually added the outer list to get concatenate to work properly. If it were just a list of numpy arrays, concatenating put all of the values into one giant numpy array and lost all information of individual samples. Putting it inside another list is what kept the dimensions since np.concatenate doesn't have a keepdims arg.

src/rastervision/tagging/tasks/predict.py
```diff
@@ -101,40 +101,82 @@ def compute_probs(run_path, model, options, generator, split):
...
-                aug_probs = np.mean(model.predict(aug_batch), axis=0,
-                                    keepdims=True)
-
-                y_probs.append(aug_probs)
-        # Otherwise, predicting probs on unaugmented images by batch
-        else:
+                aug_probs = model.predict(aug_batch)
+
+                for aug_ind in range(NUM_AUG):
+                    y_probs[aug_ind].append(aug_probs[aug_ind])
```
lewfish: If you replace `aug_probs[aug_ind]` with `np.expand_dims(aug_probs[aug_ind], axis=0)` then you will have an array of shape [1, 17]. Then when you concatenate them, you'll have an array of shape [x, 17] and you won't need to enclose in [] as we were discussing.

lewfish: I guess you could do it either way though.

lewfish:
I recommend that we close this PR since it added complexity without increasing the score.

PR #97, opened 2017-08-11
Add Planet Explorer labeled ship data generator

citerana:
This PR generates 256x256 .jpg or .tif chips from a set of labeled Planet Explorer scenes based around where ships are located in the image. Then, once the chips are created, it generates a file mapping all the ships in each chip to its pixel-wise location in the image. In order to create a set of chips from the Planet Explorer labeled ships dataset, run the following command.

```
python scripts/ship_generator.py \
    --visualize true \
    --data_type jpg \
```

Ensure that the dataset is unpacked under the detection data directory before running this command. Also, make sure to create a new directory in detection data named `singapore_train` as the script assumes it exists. The chips and accompanying `.csv` of ship locations in each chip are written here. Visualization is optional and is enabled by setting the argument to `true`.


src/detection/scripts/draw_bboxes.py
```diff
@@ -0,0 +1,199 @@
+# flake8: noqa
```
lewfish: I think you committed this file by accident.

src/detection/scripts/data_extent.py
```diff
@@ -0,0 +1,121 @@
+# flake8: noqa
+
+folders = ['20170607_024038_101d', '20170607_024039_101d',
```
lewfish: This file seems more like data than code to me. It seems like it should be a json file that's included as part of the dataset. 

However, I'm not sure why this information needs to be listed explicitly. It seems like the extents could be inferred from the data files, and then directory names can be discovered using the `os` module in Python.

src/detection/README.md
```diff
@@ -56,3 +56,15 @@ python src/detection/scripts/create_ships_tf_record.py \
...
+In order to create a set of chips from the Planet Explorer labeled ships dataset,
+run the following command.
+
+```
+python scripts/ship_generator.py \
+    --visualize true \
+    --data_type jpg \
+```
+
+The two supported outputs are tifs and jpgs. Ensure that the dataset is unpacked under the detection data directory before running this command. Also, make sure to create a new directory in detection data named `singapore_train` as the script assumes it exists. The chips and accompanying `.csv` of ship locations in each chip are written here.
```
lewfish: It would be good to explicitly state the path to the 'detection data directory' Also, your script should create the `singapore_train` directory automatically. You want to use `os.path.makedirs` with the os `exists_ok` flag.

src/detection/scripts/ship_generator.py
```diff
@@ -0,0 +1,334 @@
...
+    csv_labels = ['filename', 'xmin', 'xmax', 'ymin', 'ymax', 'class_name']
+    df = pd.DataFrame.from_records(chip_ships_list, columns=csv_labels)
+    csv_write_path = write_path + 'ship_locations.csv'
+    df.to_csv(csv_write_path, index=False)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--visualize')
+    parser.add_argument('--output_type')
```
lewfish: Good idea making these command line arguments.

src/detection/README.md
lewfish: You should also include a link to the dataset on S3.

src/detection/scripts/ship_generator.py
```diff
@@ -0,0 +1,334 @@
...
+        x[0][1]
+    ]
+
+
+def coordinate_width_and_height(corners):
+    return corners[1]-corners[0], corners[3]-corners[2]
+
+
+def points_from_rectangle(rectangle, maxXcoord, minYcoord):
+    rectangle_js = rectangle['geometry']['coordinates'][0]
```
lewfish: I'm not sure what the `_js` is supposed to mean.

src/detection/scripts/ship_generator.py
```diff
@@ -0,0 +1,334 @@
...
+    return [
+        min(maxXcoord - rectangle_js[0][0], maxXcoord - rectangle_js[1][0]),
+        max(maxXcoord - rectangle_js[0][0], maxXcoord - rectangle_js[1][0]),
+        min(rectangle_js[0][1] - minYcoord, rectangle_js[2][1] - minYcoord),
+        max(rectangle_js[0][1] - minYcoord, rectangle_js[2][1] - minYcoord)
```
lewfish: I'm not sure what operation you're trying to perform here, but if you are trying to ensure that the values don't fall outside a range, then you might want to use `np.clip`

src/detection/scripts/ship_generator.py
```diff
@@ -0,0 +1,334 @@
...
+    ]
+
+
+def scale_to_pixel(rect, height_scale, width_scale):
+    pixels = []
+    pixels.append(rect[0]*width_scale)
```
lewfish: Should have spaces around operators like `*`

src/detection/scripts/ship_generator.py
lewfish: Oh, I think it's `json`. Instead of this (which you do in some other functions), just pass in the minimal structure you need (what you are now calling `rectangle_js`) and call it `rectangle`.

src/detection/scripts/ship_generator.py
```diff
@@ -0,0 +1,334 @@
...
+
+def expand_window_with_offset(bbox):
+    """
+    This commented out portion is another method of generating offset,
+    with the new window location covering a larger area
+    box_w = abs(bbox[0] - bbox[1])
+    box_h = abs(bbox[2] - bbox[3])
+    center_x = (bbox[0] + bbox[1])/2
+    center_y = (bbox[2] + bbox[3])/2
+    valid_w = 512 - box_w
```
lewfish: These constants, like 512 and 128 should be extracted to variables so we can know what they mean and can be adjusted more easily.

src/detection/scripts/ship_generator.py
```diff
@@ -0,0 +1,334 @@
...
+                    ships_in_chip.append(relative_ship)
+        if valid_window:
+            windows.append(chip)
+            ships.append(ships_in_chip)
+
+    return windows, ships
+
+
+def generate_chips(data_type, viz):
+    from sklearn.preprocessing import MinMaxScaler
```
lewfish: I think this should be imported along with the other imports.

src/detection/scripts/ship_generator.py
```diff
@@ -0,0 +1,334 @@
...
+    datasets_path = '/opt/Data/datasets/detection'
```
lewfish: It turns out (suprise to me) that Unix isn't case sensitive, but to be consistent with the rest of the codebase, `/opt/Data` should be `/opt/data`.

src/detection/scripts/ship_generator.py
```diff
@@ -0,0 +1,334 @@
...
+    chip_ships_list = []
+    chip_ind = 0
+
+    for img_ind, folder_name in enumerate(folders):
+        folder_path = join(singapore_path, folder_name) + '/'
+        img_name = data[img_ind][0]
+        img_corners = data[img_ind][1]
+        # src_path = folder_path + img_name                       # direct data
+        src_path = folder_path + 'test_' + str(img_ind) + '.tif'  # warped data
+        ogr_path = folder_path + 'ships_ogr_' + str(img_ind) + '.geojson'
```
lewfish: You should use `os.path.join` when building up paths instead of string concatenation.

src/detection/scripts/ship_generator.py
```diff
@@ -0,0 +1,334 @@
...
+                        dst.write(src_ds.read(1, window=mask), indexes=1)  # r
+                        dst.write(src_ds.read(2, window=mask), indexes=2)  # g
+                        dst.write(src_ds.read(3, window=mask), indexes=3)  # b
+                else:
+                    with rasterio.open(
+                            write_path + chip + '.jpg', 'w',
+                            driver='JPEG', width=256, height=256, count=3,
+                            dtype='uint8') as dst:
+                        mask_img = src_ds.read([3, 2, 1], window=mask)
+                        rescaleIMG = np.reshape(mask_img, (-1, 1))
```
lewfish: Not sure why we have to read the channels in reverse order for JPGs.

src/detection/scripts/ship_generator.py
```diff
@@ -0,0 +1,334 @@
...
+                        scaler = MinMaxScaler(feature_range=(0, 255))
+                        rescaleIMG = scaler.fit_transform(rescaleIMG)
```
lewfish: `rescaleIMG` -> `rescale_img`

lewfish:
I added this feedback so you could learn more from the experience. I'm not sure it's worth making these changes though since we will probably write a more general version of this script that works with the labels generated by the Raster Foundry UI soonish.

lewfish:
There are a few changes I think you should make so that I can run this. 
* There's a bunch of dead code (ie. code that is never executed) and related imports that should be removed. In particular, I think we don't need shapely or pyproj. 
* The script needs to run inside the Docker container. Currently, we can't do that because the visualize function tries to open a window using QT and we can't do that inside the container. So, you should write the visualization chips to disk, like we do in other parts of the project. Also, there's a special way you need to import matplotlib, which you can see in other parts of the project.
* We need to generate a training and validation set, where training chips are generated from a subset of the TIFF files. This way I can make predictions on the validation TIFFs.

lewfish:
This is superseded by https://github.com/azavea/raster-vision/pull/100

PR #123, opened 2017-09-22
Exit train_ec2.sh when tensorboard quits

### LINKED ISSUE #107: Terminating AWS Batch jobs broken
If you terminate an AWS Batch job via the console or AWS CLI, it doesn't work. This happens when running the `train_ec2.sh` script. The only way to kill it is to kill the underlying spot instance. We think this is due to a bug in Batch, and should submit a bug report to AWS.

tnation14:
Move the bash signal traps in `train_ec2.sh` so that the script kills all background processes when `tensorboard` exits or the container recieves a `SIGTERM`/`SIGINT` signal.

Fixes #107.

# Testing
- Start a training job using `batch_submit.py`, then cancel it through the console.
```bash
src/detection/scripts/batch_submit.py  feature/tnation/batch-job-termination     "/opt/src/detection/scripts/train_ec2.sh \
    --config-path /opt/src/detection/configs/ships/neg/ssd_mobilenet_v1.config \
    --train-id tnation/ships2 \
    --dataset-id singapore_ships_chips_neg \
    --model-id ssd_mobilenet_v1_coco_11_06_2017"   --attempts=2
```

- See [cancelled Batch Job output](https://console.aws.amazon.com/batch/home?region=us-east-1#/jobs/queue/arn:aws:batch:us-east-1:279682201306:job-queue~2Fraster-vision-gpu/state/FAILED/job/1270a9e6-fc6a-413e-95ae-6998436407a4/queue/arn:aws:batch:us-east-1:279682201306:job-queue~2Fraster-vision-gpu/state/FAILED)

### DIFF
--- a/src/detection/scripts/train_ec2.sh
+++ b/src/detection/scripts/train_ec2.sh
@@ -53,7 +53,7 @@ cd /opt/src/detection
 
 # sync results of previous run just in case it crashed in the middle of running
 if [ "$LOCAL" = false ] ; then
-    rm -R ${LOCAL_TRAIN}/${TRAIN_ID}
+    rm -Rf ${LOCAL_TRAIN}/${TRAIN_ID}
     aws s3 sync ${S3_TRAIN}/${TRAIN_ID} ${LOCAL_TRAIN}/${TRAIN_ID}
 
     # download pre-trained model (to use as starting point) and unzip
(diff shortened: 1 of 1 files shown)


lewfish:
That's interesting that moving the trap line earlier fixes the problem. I'm not sure if we want to run train and eval in dependent Batch jobs since they need to run concurrently. BTW, I'm about to convert this script to a Python script.

tnation14:
OK, in that case I'm going to close this PR.

PR #190, opened 2018-03-14
Add docs

lewfish:
This PR adds more inline docs.


lewfish:
More documentation will be forthcoming in small chunks over the next few weeks.

PR #194, opened 2018-03-20
Add classification task skeleton

lewfish:
This PR adds a skeleton for the classification MLTask. It also changes some terminology to be more consistent with other projects (annotation -> label and label->class).


src/rv2/core/labels.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class Labels(ABC):
+    '''A set of spatially referenced labels.
+
+    A set of labels associated with their location predicted by a model or
+    provided by human labelers for the sake of training. An example of an label
+    is a bounding box in the case of object detection or a raster in case of
+    segmentation.
+    '''
```
lossyrob: Just to be clear: for classification, the classes are still spatially referenced, by tying the chip extent to the classification(s)?

src/rv2/ml_backends/keras_classification.py
```diff
@@ -0,0 +1,13 @@
+from rv2.core.ml_backend import MLBackend
+
+
+class KerasClassification(MLBackend):
```
lossyrob: Are Backends many to one to Tasks? i.e. there wouldn't be a generic "Keras" backend that could do multiple tasks, that a Backend only can perform one task?

src/rv2/core/labels.py
lewfish: Yes. I'll clarify this in the comment.

src/rv2/ml_backends/keras_classification.py
lewfish: Yes.

lewfish: Updating comments.

lossyrob:
👍 lgtm - one comment is that separating out the renames with the skeleton would have been useful, as to have a reference PR for adding the necessary pieces for adding a backend and task. Might be more effort than worth to do now, but something to think of in a future PR/set of PRs.

lewfish:
> one comment is that separating out the renames with the skeleton would have been useful, as to have a reference PR for adding the necessary pieces for adding a backend and task. Might be more effort than worth to do now, but something to think of in a future PR/set of PRs.

True, but it's in its own commit which should still be helpful.

Issue #205, opened 2018-03-23
Add Classification MLTask


lewfish:
- [x] process_training_data
- [x] train
- [x] predict
- [x] eval
- [x] test full workflow remotely
- [ ] tutorial

lewfish:
The only thing left is tutorial which is covered by #277, so I'm closing this.

PR #211, opened 2018-03-30
Support eval for classification and run full workflow

lewfish:
This PR adds support for the eval command for classification, adds a full workflow for training a classification model on cowc-potsdam, and fixes bugs that were discovered when running the full workflow.

Connects #205 

#### Testing
* Run the test workflow with the following and make sure it doesn't crash.
```
python -m rastervision.utils.chain_workflow \
    /opt/data/lf-dev/workflow-configs/classification/cowc-potsdam-test.json \
    --run
```
* Swap in the model I trained using the full workflow which is at https://github.com/azavea/raster-vision-data/releases/download/v0.0.2/cowc-potsdam-classification-model.zip, and then run the `predict` and `eval` steps again for the test workflow. The predictions viewed in QGIS (filter GeoJSON so `class_id=1`) should look like this
test-tetris
* Run the full workflow remotely using the following. This will take a long time, so you probably want to skip this and just check the results on S3.  The average F1 score was 0.97.
```
python -m rastervision.utils.chain_workflow \
    /opt/data/lf-dev/workflow-configs/classification/cowc-potsdam.json \
    --run --remote --branch lf/cleval
```

Here are the predictions on the full test file 2-13.
tetris

### DIFF
--- a/src/keras_classification/utils.py
+++ b/src/keras_classification/utils.py
@@ -28,3 +28,10 @@ def make_dir(path, check_empty=False, force_empty=False, use_dirname=False):
     if check_empty and not is_empty:
         raise ValueError(
             '{} needs to be an empty directory!'.format(directory))
+
+
+def predict(batch, model):
+    # Apply same transform to input as when training.
+    # TODO be able to configure this transform and the one in the
+    # training generator.
+    return model.predict(batch / 255.0)
--- a/src/keras_classification/commands/train.py
+++ b/src/keras_classification/commands/train.py
@@ -1,6 +1,6 @@
 import click
 
-from keras_classification.utils.files import load_json_config
+from keras_classification.utils import load_json_config
 from keras_classification.builders import trainer_builder, model_builder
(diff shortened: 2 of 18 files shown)


src/keras_classification/utils.py
```diff
@@ -28,3 +28,8 @@ def make_dir(path, check_empty=False, force_empty=False, use_dirname=False):
     if check_empty and not is_empty:
         raise ValueError(
             '{} needs to be an empty directory!'.format(directory))
+
+
+def predict(batch, model):
+    # Apply same transform to input as when training.
+    return model.predict(batch / 255.0)
```
lossyrob: Should this be configurable? Could be a TODO, but as it currently sits keras classification won't work on non-UByte imagery.

lewfish: Yea, this should be configurable. I think the transform should be done at the level of RV though, and that's a more complicated refactor since I would also need to turn off the transform being done by the TF Object Detection API. I made a TODO in the code and added a task to the backlog.

PR #215, opened 2018-04-03
Add Travis CI build; publish containers to quay.io

lossyrob:
Builds and publishes containers to quay.io via Travis.

Changes to Dockerfiles are to prevent conda from using a progress bar, which blows up Travis logs.

TODO:
- [x] Build containers for all branches; publish only on develop
- [x] Document the tag format of the docker repo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+sudo: required
+
+language:
+   - python
+
+services:
+   - docker
+
+env:
+   global:
+      - CLEAN_TRAVIS_TAG=${TRAVIS_TAG/[[:space:]]/}
+      - COMMIT=${CLEAN_TRAVIS_TAG:-${TRAVIS_COMMIT:0:7}}
+
+script:
+   - .travis/build
+
+after_success:
+   - if [ -n "$TRAVIS_TAG" ]; then
+       .travis/build "BUILD_ALL" && \
+         .travis/publish $TRAVIS_TAG "FALSE";
+     elif [ "$TRAVIS_PULL_REQUEST" == "false" -a "$QUAY_USERNAME" != "" -a "$QUAY_PASSWORD" != "" ]; then
+       .travis/build "BUILD_ALL" && \
+         .travis/publish $COMMIT "TRUE";
+     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ script:
    - .travis/build
 
 after_success:
-   - if [ -n "$TRAVIS_TAG" ]; then
+   - if [ -n "$TRAVIS_TAG" -a "$QUAY_USERNAME" != "" -a "$QUAY_PASSWORD" != ""]; then
        .travis/build "BUILD_ALL" && \
          .travis/publish $TRAVIS_TAG "FALSE";
      elif [ "$TRAVIS_PULL_REQUEST" == "false" -a "$QUAY_USERNAME" != "" -a "$QUAY_PASSWORD" != "" ]; then
--- a/.travis/publish
+++ b/.travis/publish
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+source .travis/env
+
+TAG_SUFFIX=${1}
+IS_LATEST=${2}
+
+docker login -u="$QUAY_USERNAME" -p="$QUAY_PASSWORD" quay.io
+
+function tag_and_push() {
+    docker tag ${CPU_IMAGE} "${CPU_TAG}-${1}";
+    docker tag ${GPU_IMAGE} "${GPU_TAG}-${1}";
+
+    docker push "${CPU_TAG}-${1}"
+    docker push "${GPU_TAG}-${1}"
+}
+
+tag_and_push ${TAG_SUFFIX}
+
+if [ "$IS_LATEST" == "TRUE" ]; then
+    tag_and_push "latest"
+fi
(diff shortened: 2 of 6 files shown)


lossyrob:
@lewfish this should be ready for review. There might be some issues with how it builds the develop branch or tags, but I won't be able to debug that without merging.

lossyrob:
Might be effected by #231, I can update and merge after that PR if need be

scripts/jupyter
```diff
@@ -18,17 +18,35 @@ Run the jupyter notebook in a raster-vision docker image locally
...
-           -p 8888:8888 \
-           raster-vision-cpu \
-           /run_jupyter.sh \
-           --ip 0.0.0.0 \
-           --port 8888 \
-           --no-browser \
-           --allow-root \
-           --notebook-dir=/opt/src
+    if [ -n ${RASTER_VISION_NOTEBOOK_DIR} ]; then
+        NOTEBOOK_DIR=$RASTER_VISION_NOTEBOOK_DIR
```
lewfish: It looks like `NOTEBOOK_DIR` isn't being used anywhere. I don't think you need this line.

.travis/publish
```diff
@@ -0,0 +1,22 @@
...
+
+source .travis/env
+
+TAG_SUFFIX=${1}
+IS_LATEST=${2}
+
+docker login -u="$QUAY_USERNAME" -p="$QUAY_PASSWORD" quay.io
+
+function tag_and_push() {
+    docker tag ${CPU_IMAGE} "${CPU_TAG}-${1}";
```
lewfish: Should use ${TAG_SUFFIX} instead of ${1} here and on next few lines.

lewfish:
How does this interact with the `cipublish` script? I assume we still want to use ECR since it's faster when pulling images on EC2. Maybe we should rename that script `ecr_publish`?

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+
+env:
+   global:
+      - CLEAN_TRAVIS_TAG=${TRAVIS_TAG/[[:space:]]/}
+      - COMMIT=${CLEAN_TRAVIS_TAG:-${TRAVIS_COMMIT:0:7}}
+
+script:
+   - .travis/build
+
+after_success:
```
lewfish: Why use `after_success` and not `deploy` here?

README.md
```diff
@@ -32,6 +33,17 @@ You can find more information and talk to developers (let us know what you're wo
...
 Similarly, an outdated prototype of object detection can be found at [this tag](https://github.com/azavea/raster-vision/releases/tag/old-object-detection) under the `rv` module.
 
+### Docker images
+
+Raster Vision is publishing docker images to [quay.io](https://quay.io/repository/azavea/raster-vision).
+The tag for the `raster-vision` image determines what type of image it is:
+- The `cpu-*` tags are for running the CPU containers.
+- The `gpu-*` tags are for running the GPU containers.
+
+We publish a new tag per commit to `develop`, which is tagged with the commit message.
```
lewfish: Seems like overkill to publish a tag for each commit -- once per PR seems like enough. Also, we should probably include the RV source code in the Docker image which we're not currently doing.

lewfish:
Connects #https://github.com/azavea/raster-vision/issues/257

README.md
lossyrob: It's per merge commit, i.e. merge of a PR. I'll change the language.

Including the source is not how RV currently operates - the task checks out the specific branch that you specify with the --remote flag. Are you suggesting to change the behavior? That might be better in a subsequent PR.

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+   - if [ -n "$TRAVIS_TAG" ]; then
+       .travis/build "BUILD_ALL" && \
+         .travis/publish $TRAVIS_TAG "FALSE";
+     elif [ "$TRAVIS_PULL_REQUEST" == "false" -a "$QUAY_USERNAME" != "" -a "$QUAY_PASSWORD" != "" ]; then
```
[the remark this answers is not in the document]
lossyrob: Good call, it should be in the above case as well.

.travis.yml
lossyrob: I based it off of an example that used `after_success`. Is there a benefit to using `deploy` instead?

lossyrob: Seems like `deploy` is specifically for deploying to something like heroku. https://docs.travis-ci.com/user/customizing-the-build#Deploying-your-Code

I'm not sure it works against an unsupported provider, which quay.io is. List is here: https://docs.travis-ci.com/user/deployment/#Supported-Providers

scripts/jupyter
lossyrob: Good call, I'll remove.

lossyrob:
@lewfish I think we should probably rename @cipublish, since that script is not participating in CI.

lossyrob:
This PR is tested by travis kicking off for this PR; I also [published some containers]() with a modified version of the .travis.yml script (which published this branch instead of develop)(https://quay.io/repository/azavea/raster-vision/image/d30d16e368796cf70655092a0ccf1e49b18200c6e7e04e5987e27d92ab657550). 

Testing that the develop logic is difficult without merging to develop, so I suggest we merge and fix any problems after the fact that Travis comes up with.

README.md
lewfish: I still think we should mount a specific commit of RV to the container, but that maybe the image should have RV baked into it as well so that RV can be run in the image without having to mount a volume containing the code. It depends on whether the image is supposed to be self-contained, or if it's just supposed to have the dependencies for RV. It'd be an easy thing to add later.

PR #217, opened 2018-04-11
Support for new windowing methods

dustymugs:
In addition to the standard chipping for ssd models, new windowing methods are useful for faster rcnn (or models that maintain aspect ratios).

I can't say I'm a fan of my changes to `src/rastervision/utils/files.py` to be lazy with rasters in S3 (particularly AWS NAIP S3 bucket). But it convenient instead of having to make sure that the `requester pays` flag is set...

Also had to add `scripts/compile` as it isn't in the repo


lossyrob:
Thanks for the PR! @lewfish and I will review next week.

src/rastervision/utils/files.py
```diff
@@ -56,6 +57,22 @@ def sync_dir(src_dir, dest_uri, delete=False):
...
 
 
+def _is_raster(uri, s3_test=False):
+
+    if s3_test:
+        uri = uri.replace('s3://', '/vsis3/')
+
+    try:
+
+        rasterio.open(uri)
```
lossyrob: My understanding is that the more recent versions of rasterio use the `s3://` prefix - are we using an old version?

src/rastervision/utils/files.py
```diff
@@ -56,6 +57,22 @@ def sync_dir(src_dir, dest_uri, delete=False):
...
+
+    except:
+
+        return False
+
```
lossyrob: Format nitpick: Spacing should match other code blocks.

src/rastervision/utils/files.py
dustymugs: I'm not sure about rastervision's docker images as I build my own. Unless you're forcefully installing 1.0a12 (or later), you're only on 0.3.

src/rastervision/utils/files.py
dustymugs: oops.

src/rastervision/utils/files.py
lossyrob: OK, we'll test this as part of running through the PR.

lossyrob:
Is there a sample workflow that we could use to run through this PR on a local container using public data?

dustymugs:
Yep. I made a public example available on Granular's fork. Only key item here is that the AWS NAIP bucket is requester pays so gdal/rasterio will want the following env variable: `AWS_REQUEST_PAYER=requester`

https://github.com/granularag/raster-vision/tree/master/examples/projects/pivots

lewfish:
There is already a `compile` script hiding in `src/scripts/compile`. You might still want to copy over the block you added that lets the script run from arbitrary locations. It's confusing that we have two scripts folders. We'll fix that along with some other Ops changes we're planning to make.

dustymugs:
@lewfish Ah! I did not consider `src/scripts/compile`. I'll remove mine minus that one code block

dustymugs:
@lewfish Double yes. yes because some models (published or otherwise) allow arbitrary sized input. Also, the problem is that some input data just doesn't rescale well or when simplistically chipped (for NAIP scenes in western Nebraska, each full pivot is naively 600 - 1500 pixels square (at 0.6 - 1m) and there is so much variability within that pivot.

Yes, I've wondered the same regarding prediction time detection. I am exploring random window around an annotation that is the same size of the annotation...

lewfish:
Another potential issue with using arbitrary sized inputs: it doesn't seem possible to create minibatches of size > 1 out of arbitrary sized inputs.

src/rastervision/utils/files.py
```diff
@@ -66,6 +77,12 @@ def download_if_needed(uri, download_dir, must_exist=True):
 
     parsed_uri = urlparse(uri)
     if parsed_uri.scheme == 's3':
+
+        # is it a raster?
+        vsis3_path = _is_raster(uri, s3_test=True)
+        if vsis3_path:
+            return vsis3_path
+
```
lewfish: Nice. Now we don't have download rasters unnecessarily.

lewfish:
> Another potential issue with using arbitrary sized inputs: it doesn't seem possible to create minibatches of size > 1 out of arbitrary sized inputs.

I just tried training with a batch size of 8 and didn't have any problems, so I guess this is possible.

dustymugs:
Well, that's good to know. I've been using faster rcnn models keeping batch size of 1...

src/rastervision/utils/files.py
```diff
@@ -56,6 +57,16 @@ def sync_dir(src_dir, dest_uri, delete=False):
     subprocess.run(command)
 
 
+def _is_raster(uri, s3_test=False):
+    if s3_test:
+        uri = uri.replace('s3://', '/vsis3/')
```
lewfish: I'm not sure why `s3_test` is a parameter. Shouldn't we always do this replacement?

src/rastervision/labels/object_detection_labels.py
```diff
@@ -84,6 +84,16 @@ def get_subwindow(self, window, ioa_thresh=1.0):
     def get_boxes(self):
         return [Box.from_npbox(npbox) for npbox in self.boxlist.get()]
 
+    def get_intersection_boxes(self, other):
```
lewfish: I think you added this because you want to get the boxes in the global frame of reference and not the window/local frame of reference which is what `get_subwindow` does. I think it would be helpful to add a comment or PyDoc distinguishing what this does from `get_subwindow`.

dustymugs: can do

src/rastervision/utils/files.py
dustymugs: yes

lewfish:
I'd like to get this merged today so I'm going to make a few changes.

PR #218, opened 2018-04-17
Allow specification of object detection scripts

dustymugs:
This adds support for calling externally placed tensorflow models, specifically object detection

Example...

```
...
    "train_options": {
        "pretrained_model_uri": "{rv_root}/pretrained-models/tf-object-detection-api/faster_rcnn_inception_v2_coco_2018_01_28.tar.gz",
        "backend_config_uri": "{rv_root}/backend-configs/tf-object-detection-api/faster_rcnn_inception_v2_coco.config",
        "sync_interval": 600,
        "object_detection_options": {
            "train_py": "/opt/tf_models/research/object_detection/train.py",
            "eval_py": "/opt/tf_models/research/object_detection/eval.py",
            "export_py": "/opt/tf_models/research/object_detection/export_inference_graph.py"
        }
    },
...
```

### DIFF
--- a/src/rastervision/ml_backends/tf_object_detection_api.py
+++ b/src/rastervision/ml_backends/tf_object_detection_api.py
@@ -149,18 +149,21 @@ def terminate_at_exit(process):
     atexit.register(terminate)
 
 
-def train(config_path, output_dir):
+def train(config_path, output_dir, train_py=None, eval_py=None):
     output_train_dir = join(output_dir, 'train')
     output_eval_dir = join(output_dir, 'eval')
 
+    train_py = train_py or '/opt/src/tf/object_detection/train.py'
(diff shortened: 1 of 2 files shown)


dustymugs:
Oops. Didn't mean to delete branch...

lewfish:
It seems like this is to makes it easy to switch to a different version of the TF Object Detection API which isn't pip installable. We've also been thinking about how to avoid including the TF code directly in this repo, and considered adding a script that would download a particular version of it. I'm ok with this way of doing it for now.

dustymugs:
Correct. Our docker images always use TF models outside of RV just so that we can experiment. I have a gist that generally covers what we do.

https://gist.github.com/dustymugs/d53519bbcb3fbe30f89faa8dff3f2086

PR #219, opened 2018-04-17
Add support for projects with just raster files, no annotations

dustymugs:
Depending on the training data, it really isn't feasible to build sufficient negative data for training. Instead, allow configs to pass in rasters to be used as negative data...

```
...
{
    "raster_source": {
        "geotiff_files": {
            "uris": ["s3://aws-naip/ca/2012/1m/rgbir/39122/m_3912232_sw_10_1_20120705.tif"]
        }
    },
    "ground_truth_label_store": {
        "object_detection_geojson_file": {}
    }
}
```
I'm kind of concerned with the verbosity for essentially an empty `ground_truth_label_store` but minimal changes needed to support the end goal...

### DIFF
--- a/src/rastervision/ml_tasks/object_detection.py
+++ b/src/rastervision/ml_tasks/object_detection.py
@@ -59,6 +59,7 @@ class ObjectDetection(MLTask):
         # Make positive windows which contain labels.
         pos_windows = make_pos_windows(
             raster_source.get_extent(), label_store, options.chip_size)
+        nb_pos_windows = len(pos_windows)
 
         # Make negative windows which do not contain labels.
         # Generate randow windows and save the ones that don't contain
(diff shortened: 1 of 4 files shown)


lewfish:
In order to actually generate negative chips from projects with no labels I think you'll need to make an additional change here since `len(pos_windows)` will be zero https://github.com/azavea/raster-vision/blob/develop/src/rastervision/ml_tasks/object_detection.py#L68

lewfish:
Otherwise this looks good to me.

dustymugs:
Oops. I'm missing a commit here. Let me go find it from my master branch...

src/rastervision/ml_tasks/object_detection.py
```diff
@@ -65,8 +65,11 @@ def get_train_windows(self, project, options):
         # any labels. It may take many attempts to generate a single
         # negative window, and could get into an infinite loop in some cases,
         # so we cap the number of attempts.
-        nb_neg_windows = \
-            int(options.object_detection_options.neg_ratio * len(pos_windows))
+        if nb_pos_windows:
```
lewfish: This needs to be preceded by `nb_pos_windows = len(pos_windows)`

dustymugs: Yes... you're correct...

lewfish:
This will be good to go after cleaning up the commits.

PR #220, opened 2018-04-18
Better memory management when building training data

dustymugs:
My training set results in a 24+ GB TFRecord and given that RV currently stores all chips in memory until all Training and Evaluation datasets are process, i reliably get an OOM error. Instead, write a TFRecord for each project and after all projects are done, merge all TFRecords.

This will also allow the next step of parallel building of training datasets. It takes me 2 hours right now to build the training set...

### DIFF
--- a/src/rastervision/core/ml_backend.py
+++ b/src/rastervision/core/ml_backend.py
@@ -9,9 +9,21 @@ class MLBackend(ABC):
     """
 
     @abstractmethod
-    def convert_training_data(self, training_data, validation_data, class_map,
-                              options):
-        """Convert training data to backend-specific format and save it.
+    def per_project_data_processor(self, project, data, class_map, options):
+        """Process each project's training data
+
+        Args:
+            project: Project
+            data: TrainingData
+            class_map: ClassMap
+            options: ProcessTrainingDataConfig.Options
+        """
+        pass
+
+    @abstractmethod
+    def all_projects_dataset_processor(self, training_data, validation_data,
+                                       class_map, options):
+        """After all projects have been processed, process the resultset
 
         Args:
             training_data: TrainingData
--- a/src/rastervision/ml_backends/tf_object_detection_api.py
+++ b/src/rastervision/ml_backends/tf_object_detection_api.py
@@ -75,6 +75,17 @@ def write_tf_record(tf_examples, output_path):
     writer.close()
 
 
+def merge_tf_records(output_path, src_records):
+
+    writer = tf.python_io.TFRecordWriter(output_path)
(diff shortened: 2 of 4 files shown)


dustymugs:
One thing I find to be of value in the future is that given each project having a TFRecord, this allows resume upon failure when building training data. Another interesting consideration is to build varying aggregate TFRecords given a set of project TFRecords.

src/rastervision/core/ml_backend.py
```diff
@@ -9,9 +9,21 @@ class MLBackend(ABC):
...
+            data: TrainingData
+            class_map: ClassMap
+            options: ProcessTrainingDataConfig.Options
+        """
+        pass
+
+    @abstractmethod
+    def all_projects_dataset_processor(self, training_data, validation_data,
+                                       class_map, options):
+        """After all projects have been processed, process the resultset
```
lewfish: The name `training_data` is confusing since that name is used in other places to refer to objects of type `TrainingData`. In this case, it's used to refer to some backend-specific way of representing paths to project datasets. I can see that there's no good concise way to encapsulate this in a variable name, but I think it would be helpful to explain the contents of this variable in the Pydocs, both for the abstract methods and the implementations in the backend classes.

src/rastervision/ml_backends/tf_object_detection_api.py
```diff
@@ -327,16 +337,28 @@ class TFObjectDetectionAPI(MLBackend):
...
 
-    def convert_training_data(self, training_data, validation_data, class_map,
-                              options):
+    def per_project_data_processor(self, project, data, class_map, options):
+
         training_package = TrainingPackage(options.output_uri)
+        tf_examples = make_tf_examples(data, class_map)
+        record_path = training_package.get_local_path(
+            training_package.get_record_uri(project.id))
+        write_tf_record(tf_examples, record_path)
```
lewfish: When this function exits, `training_package` is deleted which deletes the temporary directory it created, which deletes `record_path`. This is a problem because when `all_projects_dataset_processor` is executed, the paths in `training_data` no longer exist.

dustymugs: Ah! I've never tried output_uri being remote... I'll have to experiment...

lewfish:
I like the changes you made. Of course it's fine to have multiple commits for a PR, but we like to avoid commits that are just small fixes of other, more meaningful commits on the develop/master branch. I'll merge this after you squash commits.

dustymugs:
sorry for the delay. didn't see the last comment about squashing commits and my brain's been elsewhere.

PR #225, opened 2018-04-19
Temporary Directory handling

dustymugs:
* Consider the environment when setting temporary directory. In my AWS Batch envs, I explicitly provide sufficient disk space to be provided to Docker containers.
* Given that there may be more than one RasterVision process running on a host (across docker containers or directly on the host), make sure to provision a one-off temporary directory for the process.

Eventually, we should have a unified "init" mechanism to contain the per-process environment (temp dir, s3 object)

### DIFF
--- a/src/rastervision/utils/files.py
+++ b/src/rastervision/utils/files.py
@@ -1,18 +1,18 @@
-import os
-import shutil
-from urllib.parse import urlparse
-import subprocess
 import io
+import os
+from pathlib import Path
+import rasterio
+import shutil
+import subprocess
+import tempfile
 from threading import Timer
+from urllib.parse import urlparse
 
 import boto3
 import botocore
 from google.protobuf import json_format
 
 s3 = boto3.client('s3')
-
-RV_TEMP_DIR = '/opt/data/tmp/'
-
 
 class NotFoundException(Exception):
     pass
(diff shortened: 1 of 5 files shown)


src/rastervision/utils/files.py
```diff
@@ -146,8 +144,17 @@ def _sync_dir(delete=True):
...
+RV_TEMP_DIR = '/opt/data/tmp/'
+# no explicitly set tempdir
+if (not any([k in os.environ for k in ['TMPDIR', 'TEMP', 'TMP']]) and
+    not tempfile.tempdir):
+    tempfile.tempdir = RV_TEMP_DIR
+    make_dir(RV_TEMP_DIR)
+
+# now, ensure uniqueness for this process
+# the host may be running more than one rastervision process
+RV_TEMP_DIR = tempfile.mkdtemp()
```
lewfish: My impression is that this is supposed to set `RV_TEMP_DIR` to the value of the env var `TMP` when it has been set. This does not appear to happen though in Python 3.5.1. 

```
root@7a96485183fd:/opt/src# TMP='/opt/data/testing123' python
Python 3.5.1 |Continuum Analytics, Inc.| (default, Jun 15 2016, 15:32:45)
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tempfile
>>> tempfile.mkdtemp()
'/tmp/tmpqxmmyzyz'
>>> import os
>>> os.environ['TMP']
'/opt/data/testing123'
```

dustymugs: Yes, I am using this code (and generally at the env level) throughout the data science team. As for your test case that I'm also experimenting with, yes... I see the same behavior on 3.6...

Ah... I see what's going on... evidently, Python tries to be smart about it:

https://hg.python.org/cpython/file/tip/Lib/tempfile.py#l188

The user of the process must be able to write files to the directory! Which also means the directory must exist!

lewfish: Ok. Do you want to automatically create the directory then to be more consistent with how the rest of RV works?

dustymugs: Sorry. Can you elaborate on your question? Are you asking about getting rid of the hard-coded `/opt/data/tmp/`? I'm trying to understand "automatically create the directory" phrase and the code does automatically create a temporary directory for RV's use...

lewfish: Since the directory pointed to by `TMP` must exist for `tempfile` to utilize it (as you discovered), that just raises the question of whether we should create that directory if it doesn't already exist. This would prevent any confusion (like I had, not realizing that the directory needs to exist) although it seems kind of wrong to override Python's behavior in this case.

dustymugs: That's a good question. I would say "yes" but if `shutil.mktree()` fails, leaving a message (`warnings.warn`?) and let `tempfile` do what it would default to

dustymugs: I think 4cec286 does what we want...

dustymugs:
sorry for the delay. didn't see the last comment about squashing commits and my brain's been elsewhere.

lewfish:
> sorry for the delay. didn't see the last comment about squashing commits and my brain's been elsewhere.

No problem :)

PR #265, opened 2018-05-25
Exit on failed travis commands

notthatbreezy:
I noticed after https://github.com/azavea/raster-vision/pull/215 that there was a travis [failure to push the images up](https://travis-ci.org/azavea/raster-vision/builds/382979908#L1789)

screenshot from 2018-05-25 10-41-42

and there weren't updated images in quay
screenshot from 2018-05-25 10-42-10

I don't know the exact problem, but exiting on a failed command should help debug potentially

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,10 +15,4 @@ script:
    - .travis/build
 
 after_success:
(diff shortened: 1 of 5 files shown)


notthatbreezy:
Seems like this didn't actually do anything, so might not be worth it -- failed/exited with the same error but travis still marked it as successful

PR #271, opened 2018-05-31
Use labels that cover raster source

lewfish:
This PR makes it so that that the labels for a scene are the subset of the labels in the label store that cover the raster source. This makes it easier to split a raster up and use the parts individually because it is no longer necessary to also have to split up the label file. This functionality was already partially implemented, but only worked for `make_training_chips` and not `predict`. The subsetting now occurs at the point where the label file is parsed which makes it apply to all the commands.

This was tested with a split version of the `cowc-potsdam-test`. 
```
{
    "train_scenes": [
        {
            "id": "2-10-crop1",
            "raster_source": {
                "geotiff_files": {
                    "uris": [
                        "{rv_root}/processed-data/cowc-potsdam-test-split/2-10-crop1.tif"
                    ]
                }
            },
            "ground_truth_label_store": {
                "object_detection_geojson_file": {
                    "uri": "{rv_root}/processed-data/cowc-potsdam-test-split/2-10.json"
                }
            }
        }
    ],
    "test_scenes": [
        {
            "id": "2-10-crop2",
            "raster_source": {
                "geotiff_files": {
                    "uris": [
                        "{rv_root}/processed-data/cowc-potsdam-test-split/2-10-crop2.tif"
                    ]
                }
            },
            "ground_truth_label_store": {
                "object_detection_geojson_file": {
                    "uri": "{rv_root}/processed-data/cowc-potsdam-test-split/2-10.json"
                }
            }
        }
    ],
```

### DIFF
--- a/src/rastervision/labels/object_detection_labels.py
+++ b/src/rastervision/labels/object_detection_labels.py
@@ -10,7 +10,7 @@ from rastervision.core.labels import Labels
 from rastervision.labels.utils import boxes_to_geojson
 
 
-def geojson_to_labels(geojson, crs_transformer):
+def geojson_to_labels(geojson, crs_transformer, extent):
     """Extract boxes and related info from GeoJSON file."""
     features = geojson['features']
     boxes = []
@@ -33,6 +33,7 @@ def geojson_to_labels(geojson, crs_transformer):
     class_ids = np.array(class_ids)
     scores = np.array(scores)
     labels = ObjectDetectionLabels(boxes, class_ids, scores=scores)
+    labels = labels.get_intersection(extent)
     return labels
 
 
@@ -62,8 +63,8 @@ class ObjectDetectionLabels(Labels):
             boxlist.get(), boxlist.get_field('classes'), scores)
 
     @staticmethod
-    def from_geojson(geojson, crs_transformer):
-        return geojson_to_labels(geojson, crs_transformer)
+    def from_geojson(geojson, crs_transformer, extent):
+        return geojson_to_labels(geojson, crs_transformer, extent)
 
     @staticmethod
     def make_empty():
@@ -73,6 +74,11 @@ class ObjectDetectionLabels(Labels):
         return ObjectDetectionLabels(npboxes, labels, scores)
 
     def get_subwindow(self, window, ioa_thresh=1.0):
+        """Returns boxes relative to window.
+
+        This returns the boxes that overlap enough with window, clipped to
+        the window and in relative coordinates that lie between 0 and 1.
+        """
         window_npbox = window.npbox_format()
         window_boxlist = BoxList(np.expand_dims(window_npbox, axis=0))
         boxlist = prune_non_overlapping_boxes(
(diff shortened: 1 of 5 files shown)


src/rastervision/labels/object_detection_labels.py
```diff
@@ -62,8 +63,8 @@ def from_boxlist(boxlist):
             boxlist.get(), boxlist.get_field('classes'), scores)
 
     @staticmethod
-    def from_geojson(geojson, crs_transformer):
-        return geojson_to_labels(geojson, crs_transformer)
+    def from_geojson(geojson, crs_transformer, extent):
```
lossyrob: Should `extent` be required at a higher level? It seems a bit off to require extent for GeoJSON labels, but not for object detection labels in general. Or, perhaps the geojson labels shouldn't require an extent and this functionality be solved a different way. For example, will there be times we want to read in a set of geojson labels, where we don't have an extent? Maybe not now, but I could see that happening in the future - and we don't want to require someone to make a false extent to get use out of the class. I'd be in favor of just having call sites that require the subset of labels that overlap an extent to explicitly call `labels.get_intersection(extent)`, rather then bake in the requirement of all labels (or, in this case, only a subclass of OD labels) always having an extent passed in.

src/rastervision/labels/object_detection_labels.py
```diff
@@ -84,15 +90,16 @@ def get_subwindow(self, window, ioa_thresh=1.0):
...
-    def get_intersection_boxes(self, other):
-        boxes = []
-        other_shapely = other.get_shapely()
-        for npbox in self.boxlist.get():
-            intersection = other_shapely.intersection(
-                Box.from_npbox(npbox).get_shapely())
-            if not intersection.is_empty:
-                boxes.append(Box.from_shapely(intersection))
-        return boxes
+    def get_intersection(self, window):
```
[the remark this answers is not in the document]
lewfish: Good point; will change.

src/rastervision/labels/object_detection_labels.py
lewfish: I see what you mean. But, we're already passing in the crs_transformer to the label_store (which is associated with the raster_source), so it doesn't seem like much of a stretch to also pass in the extent. Doing it this way also avoids potential future bugs where people forget to do the get_overlap operation when its needed. So I'm leaning towards leaving it as is.

lossyrob: OK. TBH, I have had the same sort of issue with having to pass in a crs_transformer everywhere - there has been occasion (though I can't remember the specific instance) where I wanted to new up an object but was required to pass in a crs_transformer, which I didn't have. This type of coupling makes it harder to utilize things in new situations as time goes on, and this would put us further down that path. Then again, I'm not that strongly against it. Let's just be on the lookout for dependency couplings that cause pain in the future, and perhaps refactor them away later.

lewfish: I see how the dependence of LabelStore on RasterSources is annoying, but I don't see how we can decouple them, since you need to know the CRS of the rasters to convert from map coordinates in the JSON files to the pixel coordinates used by the LabelStore. We can talk about this in person, but for now I'll just merge this.

Issue #274, opened 2018-06-04
Copy RV code into Docker image


lewfish:
The Docker image should contain the RV code. This way, the images created by CI will contain the RV code from that commit and end users won't have to copy it in. In the future, we may switch to a different way of specifying the code to use when running remotely (see #228), but copying in the RV code during CI shouldn't prevent or complicate any of those options.

lewfish:
As part of this, we should also run the `compile` script so end users don't need to do this themselves.

Issue #258, opened 2018-05-24
Add unit tests


lewfish:
Add some unit tests and get them to run in CI.

PR #275, opened 2018-06-05
Add unit tests for Box

lewfish:
This PR adds unit tests for the Box class and makes it so that all unit tests are run by Travis CI. It also copies RV code into the Docker images.

Connects #258
Connects #274


src/rastervision/core/box_test.py
```diff
@@ -0,0 +1,155 @@
+import unittest
```
lossyrob: A bit unfamiliar with how python unittest works...

Does it matter from a future-pip-installable form of rastervision that we are containing the code inside the src folders that contain the actual library code? Does it make sense to separate the test code out from the main source? (This question is possibly fueled by my jvm/scala-centric worldview)

lewfish: I don't think the current test layout would prevent making it pip installable, but if it did, it would be easy to change. You can put unit tests wherever you want. I like putting the tests alongside the source files because then you don't have to maintain a parallel file hierarchy for the tests. It's also easier to navigate to the tests for a file. This is how the Object Detection API does it. See https://github.com/tensorflow/models/tree/master/research/object_detection/core

lossyrob: :+1:

Issue #276, opened 2018-06-06
Make sure all dependency versions are correct


tnation14:
`develop` builds are currently failing because the pinned version of `git` (version `1:2.7.4-0ubuntu1.3`) isn't available on Ubuntu Xenial. Pin git to `1:2.7.4-0ubuntu1.4`, and make sure all other dependencies are pinned to available versions. In order to reduce the maintenance burden of keeping these pins up to date, it may be better to use wildcard notation and pin to a major release instead of a patch.

lewfish:
Thanks for doing this. I didn't know what I was doing when I pinned it to those version :)

tnation14:
Closed via #278

PR #278, opened 2018-06-06
Pin dependencies to minor versions

tnation14:
# Overview

Use wildcard notation to pin all dependencies to a minor version. This ensures that patch updates don't break builds. Additionally, enable `pipefail` behavior on `src/scripts/install_deps` so that any failed dependency installation fails the build. This ensures that the images we push up to quay have built properly.

# Testing
- See Travis build checks
- Apply the following diff, then run `.travis/build`. Ensure that the build exits with an error.
```diff
diff --git a/src/scripts/install_deps b/src/scripts/install_deps
index d6901b0..621d283 100755
--- a/src/scripts/install_deps
+++ b/src/scripts/install_deps
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -ex
 
+exit 1
+
 apt-get update
 apt-get install -y wget=1.* git=1:2.* \
     python-protobuf=2.* python3-tk=3.*
```

# Notes
I'm open to unpinning `git` and `wget` versions, but I imagine `python-protobuf` and `python3-tk` need to stay pinned.

### DIFF
--- a/src/scripts/install_deps
+++ b/src/scripts/install_deps
@@ -2,8 +2,8 @@
 set -ex
 
 apt-get update
-apt-get install -y wget=1.17.1-1ubuntu1.4 git=1:2.7.4-0ubuntu1.3 \
-    python-protobuf=2.6.1-1.3 python3-tk=3.5.1-1
+apt-get install -y wget=1.* git=1:2.* \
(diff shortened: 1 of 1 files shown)


tnation14:
Fixes #276

PR #294, opened 2018-07-03
Clarify crs_transformer naming

lewfish:
This PR changes `web` to `map` in crs_transformer method and variable names and adds more comments. This is to reflect the fact that map-based coordinates are not necessarily in web mercator.

Closes #294

### DIFF
--- a/src/rastervision/crs_transformers/rasterio_crs_transformer.py
+++ b/src/rastervision/crs_transformers/rasterio_crs_transformer.py
@@ -4,19 +4,17 @@ from rastervision.core.crs_transformer import CRSTransformer
 
 
 class RasterioCRSTransformer(CRSTransformer):
-    """Transformer for a RasterioRasterSource.
+    """Transformer for a RasterioRasterSource."""
 
-    This assumes that the map coordinates are always in lon/lat format.
-    """
-
-    def __init__(self, image_dataset):
+    def __init__(self, image_dataset, map_crs='epsg:4326'):
         """Construct transformer.
 
         Args:
             image_dataset: Rasterio DatasetReader
+            map_crs: CRS code
         """
         self.image_dataset = image_dataset
-        self.map_proj = pyproj.Proj(init='epsg:4326')
+        self.map_proj = pyproj.Proj(init=map_crs)
         image_crs = image_dataset.crs['init']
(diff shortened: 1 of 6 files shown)


src/rastervision/crs_transformers/rasterio_crs_transformer.py
```diff
@@ -4,22 +4,48 @@
 
 
 class RasterioCRSTransformer(CRSTransformer):
+    """Transformer for a RasterioRasterSource.
+
+    This assumes that the map coordinates are always in lon/lat format.
```
lossyrob: Is there a reason to not allow for a CRS to be optionally passed in to __init__, defaulting to EPSG:4326?

src/rastervision/crs_transformers/rasterio_crs_transformer.py
```diff
@@ -4,22 +4,48 @@
...
 
-    def web_to_pixel(self, web_point):
+    def map_to_pixel(self, map_point):
+        """Transform point from map to pixel-based coordinates.
+
+        Args:
+            map_point: (long, lat) tuple
+
+        Returns:
+            (x, y) tuple in pixel coordinates
```
lossyrob: In GeoTrellis we differentiate between grid coordinates (pixel coordinates) and map coordinates  by always referring to the former as `(col, row)` and always referring to  the latter as `(x, y)`. We might want to adopt a similar strategy - it  has been useful for (somewhat effectively) preventing bugs that are based on folk not knowing in which reference  system they are working in.

src/rastervision/crs_transformers/rasterio_crs_transformer.py
lewfish: I added this.

src/rastervision/crs_transformers/rasterio_crs_transformer.py
lewfish: l will consider doing this when I write unit tests (and lightly refactor) the Labels and LabelStore classes.

lewfish:
To handle non default CRSs in GeoJSON files requires a further change but I will do that as part of https://github.com/azavea/raster-vision/issues/307

Issue #298, opened 2018-07-05
Write ADR for repeatable containerized experiments


lewfish:
Relates to #228

PR #299, opened 2018-07-06
Add ADR on Repeatable Containerized Experiments

lewfish:
Connects #298


lewfish:
I'm going to close this for now since we won't be able to do all this work before the release. After then I will revise this ADR and then implement it.

PR #302, opened 2018-07-10
support multipolygons

simonkassel:
This PR adds support for multipolygon geojson files. It converts each polygon within a feature to it's own bounding box when converting from geojson to labels.

### DIFF
--- a/src/rastervision/labels/object_detection_labels.py
+++ b/src/rastervision/labels/object_detection_labels.py
@@ -17,18 +17,25 @@ def geojson_to_labels(geojson, crs_transformer, extent):
     class_ids = []
     scores = []
 
-    for feature in features:
+    def polygon_to_label(polygon, crs_transformer):
         # Convert polygon to pixel coords and then convert to bounding box.
-        polygons = feature['geometry']['coordinates'][0]
-        for single_polygon in polygons:
-            polygon = [crs_transformer.web_to_pixel(p) for p in single_polygon]
-            xmin, ymin = np.min(polygon, axis=0)
-            xmax, ymax = np.max(polygon, axis=0)
-            boxes.append(Box(ymin, xmin, ymax, xmax))
+        polygon = [crs_transformer.web_to_pixel(p) for p in polygon]
+        xmin, ymin = np.min(polygon, axis=0)
+        xmax, ymax = np.max(polygon, axis=0)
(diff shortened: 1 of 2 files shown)


src/rastervision/labels/object_detection_labels.py
```diff
@@ -19,15 +19,16 @@ def geojson_to_labels(geojson, crs_transformer, extent):
...
-        polygon = feature['geometry']['coordinates'][0]
-        polygon = [crs_transformer.web_to_pixel(p) for p in polygon]
-        xmin, ymin = np.min(polygon, axis=0)
-        xmax, ymax = np.max(polygon, axis=0)
-        boxes.append(Box(ymin, xmin, ymax, xmax))
-
-        properties = feature['properties']
-        class_ids.append(properties['class_id'])
-        scores.append(properties.get('score', 1.0))
+        polygons = feature['geometry']['coordinates'][0]
```
lossyrob: Seems like this wouldn't handle non-multipolygons. Perhaps we check the "type" field  here:

```python
geom_type = feature['geometry']['type']
if geom_type == 'polygon':
  ...
else if geom_type == 'multipolygon':
  ...
else:
  raise  Exception("Geometries of type {} are not supported in object detection labels.".format(geom_type))
```
or something along those lines.

simonkassel:
yeah, i realized that and made a similar change but seeing your comment now I realize that just checking the geometry type would be cleaner

simonkassel:
I have tested this on both the cowc-potsdam (Polygon) and spacenet (MultiPolygon) datasets

lewfish:
I'd like us to get into the habit of writing unit tests for new features whenever we're not under time pressure. Could you please add a unit test for the `polygon_to_label` function? You can follow the example of other unit tests that are already in RV. Let me know if you need help with this.

simonkassel:
Yeah, I'm happy to do that. I'll let you know if I need help (likely will)

lewfish:
You'll want to add the new functionality around here: https://github.com/azavea/raster-vision/blob/develop/src/rastervision/label_stores/object_detection_geojson_file.py#L39

Then you'll add a new test case by adding a method similar to this one: https://github.com/azavea/raster-vision/blob/develop/src/rastervision/label_stores/object_detection_geojson_file_test.py#L111

PR #331, opened 2018-07-25
Add support for prediction packages

### LINKED ISSUE #308: Support prediction packages
After training, RV should create a "prediction package" which is a zip file containing a raster stats file, a predict command config template, and a model file. The predict command (or maybe a special version of it called `predict-package`?) should be able to take the package, a URI to a TIFF file, and a URI to the output and make predictions. This will simplify the implementation of the prediction API, the 
 distribution of pretrained RV models, and the process of making predictions on a pretrained model for new users.

lewfish:
## Overview

After training a model as part of a workflow, we often wish to deploy it somewhere and make predictions on new images. Doing this requires collecting a few files (including the predict command config file), and then (programmatically) modifying the config file to accommodate the new images and running the `predict` command using this modified config file. Since this is such a common operation, it makes sense to encapsulate it within RV, rather than making consumers of RV responsible for it. 

To do this, we introduce the concept of a prediction package, which is a zip file containing the files necessary for "cloning" a prediction job and running it on new images. The predict command now generates a prediction package when it is finished running. A new command, `predict_package` takes as command line arguments a package URI, a labels URI (for output), and a list of image URIs, and makes predictions. 

### Checklist

### Notes

* Originally I thought the `train` command (instead of `predict`) would generate the prediction package, but I realized this was somewhat problematic since the config file for `train` doesn't contain a lot of the info necessary to generate a prediction package.
* Running `predict_package` will run `predict` using the same options that were captured when generating the prediction package. We may want to add command line options to `predict_package` in the future to support overriding some of the options (such as `channel_order`).

## Testing Instructions

* Run the cowc-potsdam workflow and note that the predict output directory contains a `prediction-package.zip` file.
* Use the sample [cowc-potsdam prediction packages](https://github.com/azavea/raster-vision-data/releases/tag/v0.0.4) on a sample tiff file and check the predictions in QGIS. Example:
```
python -m rastervision.run predict_package \
    /opt/data/lf-dev/prediction-packages/cowc-potsdam-od.zip \
    /opt/data/lf-dev/labels.json \
    /opt/data/lf-dev/processed-data/cowc-potsdam-test/2-13.tif
```

Closes #308


jamesmcclain:
```
Name                                                           Stmts   Miss  Cover
----------------------------------------------------------------------------------
...
./rastervision/builders/predict_builder.py                        11     11     0%
...
./rastervision/commands/predict.py                                 8      8     0%
./rastervision/commands/predict_package.py                        14     14     0%
...
./rastervision/core/predict_package.py                            52      2    96%
...
./rastervision/protos/predict_pb2.py                              34      0   100%
...
----------------------------------------------------------------------------------
TOTAL                                                           2931   2058    30%
```

PR #348, opened 2018-08-08
Prototype of workflow library

lewfish:
## Overview

Previously, we compiled a chain workflow config file into a set of command
config files. This was difficult for beginners because it requires
understanding how the compilation process works, which also happens to be
poorly documented. It was also inflexible and did encapsulate knowledge of good
defaults. The new way of generating a chain workflow is to use a library of
utilities for programmatically creating configs, which fixes the above
limitations.

* rastervision.workflows.config_utils contains generic functions for creating
various protobuf configs
* rastervision.workflows.chain_utils contains more opinionated classes for
creating chain workflows
* rastervision.workflows.sample is a script that generates a workflow using the
utils. If you change the base_uri field and then run it, the script should
generate a chain workflow json file, which functions as the artifact/manifest
files we've discussed.

Some details are missing, but this gives an idea of the direction we are moving
in.

### Checklist

- [ ] Ran scripts/format_code and commited any changes

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

Closes #XXX


lewfish:
@lossyrob: I trimmed the sample.py file from 100+ lines when you saw it to 38 lines now.

lewfish:
@lossyrob: I created a function `make_classification_geotiff_geojson_scene` (https://github.com/azavea/raster-vision/pull/348/commits/24bdccd8dca578ae7ec133af059053637781c42a#diff-24a25c0e861ff613c626e76191ea5ddbR82) that is  specific to scenes with geotiffs and classification geojson files. It might seem more reasonable to create a function that is more generic and takes `task_type`, `label_store_type`, and `raster_type` as an argument. However, this won't work because each combination of these entails a different set of options. So is the way I'm doing it reasonable? I'm asking because I'm also going to need to make specialized functions for `make_training_chips_object_detection` and `make_training_chips_classification` since they have different sets of options.

PR #349, opened 2018-08-08
Add object detection integration test

### LINKED ISSUE #259: Add integration tests
Currently, we test the system by running the workflow on a tiny test version of the cowc-potsdam dataset and checking the predictions and eval metrics. In order to "fast forward" over the time consuming training step, after training for a few steps, we swap in a previously trained model from the raster-vision-data repo. As a first stab at writing integration tests, we should just script this process, and compare output to the expected output which is stored somewhere (in a fixtures directory?).

### LINKED ISSUE #315: Add object detection integration test
Part of #259

### LINKED ISSUE #333: Eval crashes if no predictions
If there are no predicted boxes, eval will crash.

### LINKED ISSUE #334: Box.make_random_square ignores lower bounds
The Box.make_random_square method assumes that the box's lower bounds are at 0,0. Instead, we should take into account the ymin and xmin.

lewfish:
## Overview

This PR adds an integration test that runs an object detection workflow. Similar to the classification test, it overfits a model on a synthetic scene, does eval on that same scene, and checks that the F1 score is 1.0. The test runs in ~5 minutes on my laptop and passes 5/5 times. If it fails we may need to increase the number of training steps.

This also fixes two bugs discovered along the way: #334 and #333

### Checklist

### Notes

There were a few unusual things that were required to train an object detection model in <= 5 mins and be able to memorize the data and get a perfect F1 score.
* We created a small test scene with colored squares using a synthetic scene generator to try to make it easier to learn. See `rastervision.contrib.scene_generator`.
* We added and used a new `sliding` option to `window_method` so that `make_training_chips` would use the same windows that the predict` command uses. This makes it  so the model would see the same images during training and prediction and have a chance to get to get an F1 of 1.0. Unfortunately, this doesn't exercise our usual windowing method.
* We turn off the eval and Tensorboard threads to conserve resources.
* We trained the model for 2500 steps to an F1 score of 0.9 and then use it as a pretrained model (stored at https://github.com/azavea/raster-vision-data/releases/tag/v0.0.5). This way, we give the model a head start, so it only needs to train for another 500 steps to get passed the finish line instead of the full 3000 steps.
* The default behavior of TFOD for finetuning models is to only copy the weights from the base of the model, but want to copy all weights from the pretrained model, so we had to set `load_all_detection_checkpoint_vars: true`. See https://github.com/tensorflow/models/issues/4944
* We turned off all data augmentation in `backend.config` so the model is trained on the same images it will see during prediction.

## Testing Instructions

* Change the `num_steps` in `backend.config` to 2501 and run `./scripts/integration_tests.py object-detection` locally. Check that it fails because the F1 score is < 1.0
* Check that this test runs in CI and that it passed.

Closes #315
Closes #259 
Closes #334
Closes #333


src/rastervision/ml_tasks/object_detection.py
```diff
@@ -69,6 +69,10 @@ def make_pos_windows(image_extent, label_store, options):
         return _make_label_pos_windows(image_extent, label_store, options)
     elif window_method == 'image':
         return [image_extent.make_copy()]
+    elif window_method == 'sliding':
```
lossyrob: Confused about this part - we're making positive windows out of the whole extent? Is this not really making positive windows, but just making windows that might  have no labels? how does that effect the make_neg_windows? How does this work downstream when the label store is queried for labels for these extents?

lossyrob:
Also something is messed up in travis wrt the protobuf https://travis-ci.org/azavea/raster-vision/builds/416383916#L2080

src/rastervision/ml_tasks/object_detection.py
lewfish: I can see why this is confusing. The `make_positive_windows` function with the `sliding` options is returning windows that are a mix of positive and negative. I think I can can make this clearer by just moving the code for `sliding` up a level to the `get_train_windows` method and not invoke the notions of positive and negative windows .

lewfish: As it stands, I'm just setting the `neg_ratio` to 0 so there are no extra negative windows created. There's nothing special going on with the label store -- it can be queried for arbitrary windows. It's not like the classification label store that can only be queried for grid aligned windows.

lewfish:
> Also something is messed up in travis wrt the protobuf https://travis-ci.org/azavea/raster-vision/builds/416383916#L2080

Ok, will fix.

src/rastervision/ml_tasks/object_detection.py
lossyrob: OK. Maybe there should be an issue to refactor in the future, to simplify/clarify the pos/neg window creation in this case? Might be a good starter issue for someone.

lewfish: I refactored to make this clearer, but I forgot to put the change in a separate commit. It's here https://github.com/azavea/raster-vision/pull/349/commits/e9a30127969495c96042f40efebfff00df4374d5

PR #351, opened 2018-08-15
Add nodata handling to rasterio raster source.

lossyrob:
## Overview

This small change ensures that if the incoming raster has a non-zero NODATA value that  it is replaced with a 0, which the system expects.

### Notes

We haven't written unit tests for the raster source code yet, this seemed like a small enough change that could go through; debatable.


lewfish:
Does this also handle nodata masks?

lewfish:
I can take this commit and then add a test and change the predict functionality to throw out boxes over nodata.

lewfish:
I will open another PR on this topic soon.

lossyrob:
Superseded by #361

PR #363, opened 2018-08-18
Allow users to configure pre-trained weights for Keras classification.

### LINKED ISSUE #360: Refactor keras_classification to use pretrained_model_uri
Currently, keras_classification uses the default Keras behavior for downloading pretrained models. We should refactor so it respects the RV pretrained_model_uri field. This make it easier to fine-tune models generated by RV. As part of this, we should update the config library to cache pretrained classification models on S3, the way it works for object detection models.

lossyrob:
## Overview

This PR allows the configuration of pre-trained weights for Keras classification.

Additionally, model weights are saved along side the model, so that model weights from models trained by Raster Rision can be used as pre-trained weights for other experiments in Raster Vision.

### Notes

This change causes some default behavior to change - previously, we were always using pretrained weights from `imagenet` for resnet50 classification. In order for those weights to be used, they now need to be stated in the workflow configuration, as seen in the classification sample workflows.

Additionally, Keras has two ways to load model weights: 
- "by name", in which weights are loaded for layers of the neural network only if they have they have the same name as layers in the HDF5 weights file.
- The default method, which assumes that the architectures of the neural networks and the weights to be loaded are the same.

The imagenet pretrained weights were being loaded "by name".  In order to account for this, I added a backend configuration option to the `model` section that specifies a `load_weights_by_name` parameter, which defaults to false. This means, to get the default behavior prior to this PR, you'll need to put in the appropriate pre-trained weight file in `pretrained_model_uri` in the workflow configuration, as well as change the backend to set `load_weights_by_name` to true.

One other change that is made for this PR is: previously, on re-running  the `train` command for a classification model, the model would not respect new model configurations, because the code saw the existence of the model file and loaded the model from that file. Further, because the  `log.csv`  file existed, the model would not be retrained - the training step would be completely skipped. This only affected local  rv_roots, not remote runs. A change in the code will now delete out the model and log file if training is re-run.

## Testing Instructions

* Run classification integration testss
* Run through COWC classification tutorial
* Use the `cowc-potsdam-pretrained-test.json` workflow to train a model that is already pretrained on COWC, and see the evaluation predictions are better than using pretrained imagenet.

Closes #360

### DIFF
--- a/src/rastervision/ml_backends/keras_classification.py
+++ b/src/rastervision/ml_backends/keras_classification.py
@@ -78,14 +78,31 @@ class DatasetFiles(FileGroup):
 class ModelFiles(FileGroup):
     """Utilities for files produced when calling train."""
 
-    def __init__(self, base_uri):
+    def __init__(self, base_uri, replace_model=False):
+        """Create these model files.
+
+        Args:
+            base_uri: Base URI of the model files
+            replace_model: If the model file exists, remove.
+                           Used for the training step, to retrain
+                           existing models.
+
+        Returns:
+            A new ModelFile instance.
+        """
         FileGroup.__init__(self, base_uri)
 
         self.model_uri = join(self.base_uri, 'model')
         self.log_uri = join(self.base_uri, 'log.csv')
 
-    def download_backend_config(self, backend_config_uri, dataset_files,
-                                class_map):
+        if replace_model:
+            if os.path.exists(self.model_uri):
+                os.remove(self.model_uri)
+            if os.path.exists(self.log_uri):
+                os.remove(self.log_uri)
+
+    def download_backend_config(self, pretrained_model_uri, backend_config_uri,
+                                dataset_files, class_map):
         config = load_json_config(backend_config_uri, PipelineConfig())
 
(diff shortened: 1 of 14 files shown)


src/rastervision/ml_backends/keras_classification.py
```diff
@@ -78,14 +78,31 @@ def _upload(data_uri):
 class ModelFiles(FileGroup):
     """Utilities for files produced when calling train."""
 
-    def __init__(self, base_uri):
+    def __init__(self, base_uri, replace_model=False):
```
lewfish: As a user, how does one specify that replace_model should be False, and that you want to continue training the model based on the checkpoint and log?

lossyrob: Ah, I see - so if there is a checkpoint already, and you've either increased the number of epochs to train or the training was interrupted,  then keeping the model and log will let you continue from the last train point.

I hadn't considered this, let me think through this use case and enable it.

lossyrob: Updated to add a train config option that allows the user to set this.

PR #369, opened 2018-08-23
Add batching of chips for prediction 

### LINKED ISSUE #356: Add option to use batch size > 1 for prediction
This is to make better use of the GPU. We also need to make corresponding changes to `predict_package`, and do a crude benchmark to check the speedup.

lossyrob:
## Overview

Adds the ability to batch up chips and send off to the backend  for processing. Allows the backend to more efficiently process larger sets of data, rather than sequentially going through small amounts of data.

## Testing Instructions

* Run integration  tests
* Tested against other GPU scenarios. Resulted in small speedup.

Closes #356

### DIFF
--- a/src/rastervision/core/evaluation_item.py
+++ b/src/rastervision/core/evaluation_item.py
@@ -36,7 +36,7 @@ class EvaluationItem(object):
 
             def weighted_avg(self_val, other_val):
                 if self_val is None and other_val is None:
-                    return None
+                    return 0.0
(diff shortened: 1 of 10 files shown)


lossyrob:
@jamesmcclain this should make a difference in segmentation predict, but not by much.

jamesmcclain:
Review incoming.

jamesmcclain:
It appears that the CI failure is due to the style checker, but I get no error when I run that locally.

PR #370, opened 2018-08-23
Segmentation: Evaluation

jamesmcclain:
## Overview

This adds evaluation support to the DeepLab-based segmentation code.

### Checklist

- [x] Unit tests
- [x] https://github.com/azavea/raster-vision/pull/366
- [x] https://github.com/azavea/raster-vision/pull/369
- [x] https://github.com/azavea/raster-vision/pull/371

### DIFF
--- a/src/rastervision/label_stores/segmentation_raster_file_test.py
+++ b/src/rastervision/label_stores/segmentation_raster_file_test.py
@@ -53,8 +53,7 @@ class TestSegmentationRasterFile(unittest.TestCase):
         data[4:, 4:, :] = [1, 1, 1]
         raster_source = TestingRasterSource(data=data)
         label_store = SegmentationInputRasterFile(
-            source=raster_source,
-            raster_class_map={'#010101': 1})
+            source=raster_source, raster_class_map={'#010101': 1})
         extent = Box(0, 0, 10, 10)
         self.assertTrue(label_store.enough_target_pixels(extent, 30, [1]))
 
@@ -63,8 +62,7 @@ class TestSegmentationRasterFile(unittest.TestCase):
         data[7:, 7:, :] = [1, 1, 1]
         raster_source = TestingRasterSource(data=data)
         label_store = SegmentationInputRasterFile(
-            source=raster_source,
-            raster_class_map={'#010101': 1})
+            source=raster_source, raster_class_map={'#010101': 1})
         extent = Box(0, 0, 10, 10)
         self.assertFalse(label_store.enough_target_pixels(extent, 30, [1]))
 
--- a/src/rastervision/label_stores/utils_test.py
+++ b/src/rastervision/label_stores/utils_test.py
@@ -22,7 +22,6 @@ class DoubleCRSTransformer(CRSTransformer):
 
 
 class TestLabelStoreUtils(unittest.TestCase):
-
     def setUp(self):
         self.crs_transformer = DoubleCRSTransformer()
         self.aoi_dict = {


src/rastervision/builders/label_store_builder.py
```diff
@@ -3,7 +3,7 @@
 from rastervision.label_stores.classification_geojson_file import (
     ClassificationGeoJSONFile)
 from rastervision.label_stores.segmentation_raster_file import (
-    SegmentationRasterFile)
+    SegmentationInputRasterFile, SegmentationOutputRasterFile)
```
lewfish: Interesting that you decided to split these into two classes. I think it's the case that we either read *or* write to a raster, but never both so I guess it makes sense to isolate the functionality that way. Maybe we should consider doing that (as a later refactor) to the other label stores.

src/rastervision/builders/label_store_builder.py
```diff
@@ -31,12 +31,13 @@ def build(config,
...
-        raster_class_map = config.segmentation_raster_file.raster_class_map
-        return SegmentationRasterFile(
-            src=config.segmentation_raster_file.src,
-            dst=config.segmentation_raster_file.dst,
-            raster_class_map=raster_class_map)
-        return None
+        srf = config.segmentation_raster_file
+        if not str(srf.source) == '':
+            return SegmentationInputRasterFile(
+                source=srf.source, raster_class_map=srf.raster_class_map)
```
lewfish: Now that there are separate classes, it might make sense to just replace `source` with `uri` and `sink` with `uri` since it's already clear what they are.

src/rastervision/evaluations/segmentation_evaluation.py
```diff
@@ -0,0 +1,91 @@
...
+
+        # Compute the intersection of the extents of the ground truth
+        # labels and predicted labels.
+        gt_extent = ground_truth_label_store.source.get_extent()
+        pr_extent = prediction_label_store.source.get_extent()
+        xmin = max(gt_extent.xmin, pr_extent.xmin)
+        ymin = max(gt_extent.ymin, pr_extent.ymin)
+        xmax = min(gt_extent.xmax, pr_extent.xmax)
+        ymax = min(gt_extent.ymax, pr_extent.ymax)
+        extent = Box(xmin=xmin, ymin=ymin, xmax=xmax, ymax=ymax)
```
lewfish: Consider extracting a function for computing intersections and moving it to the Box class.

src/rastervision/builders/label_store_builder.py
jamesmcclain: :+1:

src/rastervision/builders/label_store_builder.py
jamesmcclain: :+1: I will do so in a subsequent pull request

src/rastervision/evaluations/segmentation_evaluation.py
jamesmcclain: :+1: Will do

src/rastervision/builders/label_store_builder.py
jamesmcclain: Actually on second look, there are places where these different attribute names are relied-on.  This this area is the subject of the larger-scale refactor that is going on, it might not make sense to spend too much effort on this.

src/rastervision/evaluations/segmentation_evaluation.py
jamesmcclain: Done.

jamesmcclain:
All comments addressed.

PR #381, opened 2018-09-14
Major API refactor

### LINKED ISSUE #237: Override backend config options
Each backend is presumed to have its own config format for training. Currently, we handle this by referencing a backend config file from within the RV train command config and workflow config files. This can get annoying because every time you want to modify an existing workflow, you need to create a new backend config file, upload it, and reference the new file name from within the workflow config. I think it would be easier if we did away with backend config files and just inlined its contents as a string field in the RV config file. It's safe to assume that all backend config files will be strings, although we can't know in advance what the format will be. When training a model, the backend class will be responsible for writing this string to a file which will be fed into the actual backend executable. Or, if the actual backend executable takes command line args instead of a config file, the backend class can parse the config string into the appropriate command line args.

### LINKED ISSUE #256: Support tree workflows
We currently have support for chain workflows which are an abstraction over a chain (ie. series) of commands. We would like to add support for tree workflows, which are an abstraction over a branching, tree-structured set of experiments.

screen shot 2018-05-23 at 9 11 08 pm

I think we can implement this by reusing a lot of the functionality in the chain workflow runner. It's not clear how to best represent the tree / set of experiments to run. It could be represented procedurally in code using nested for loops like we've done in the past, or it could be represented more declaratively in a tree workflow config file that is like a chain workflow config, but specifies a range of values instead of a single value for certain fields.

### LINKED ISSUE #358: Support branching experiments
We'd like to support the ability to run an experiment and then create a branch off of it. In other words, you should be able to run an experiment, and then create a modification of it, perhaps just changing the predict and eval phase. We can implement this by: 
* adding ability to load an experiment that was run from its config file
* adding an update method for each command to the Experiment class. Rob's suggestion is to make Experiment immutable, so that calling each of these methods will return a new Experiment.

### LINKED ISSUE #362: Remove temporary directory in TF object detection training  package
https://github.com/azavea/raster-vision/blob/develop/src/rastervision/ml_backends/tf_object_detection_api.py#L300

This temporary directory is  not cleaned up.

### LINKED ISSUE #368: Develop API for "hot predictions"
The  prediction command (through a prediction config, chain workflow or prediction package) is good for making batch predictions. However, it's not ideal for doing predictions-as-a-service. We should have library API that allows you to initialize a object or method that allows for predictions from a loaded mode. For instance, this is what I'm doing to allow for an initialized model that can call "predict" at of a loaded model for classification and object detection:

```python
lass Model:
    def __init__(self, predict_package_path, temp_dir):
        config = load_predict_package(predict_package_path,
                                      temp_dir,
                                      "",
                                      ["NONE"], # passing in false image
                                      update_stats=False,
                                      channel_order=None)

        # Save off the generated scene config as a template, then remove scenes
        # as the command building requires processing of real scenes.
        scene_template = config.scenes[0]
        del config.scenes[:]

        # Move the stats uri out of the temp directory
        new_stats_uri = os.path.join(STATS_DIR, "{}-stats.json".format(uuid.uuid1()))
        shutil.copy(scene_template.raster_source.raster_transformer.stats_uri,
                    new_stats_uri)
        scene_template.raster_source.raster_transformer.stats_uri = new_stats_uri

        # Build command
        command = predict_builder.build(config)

        self.command, self.scene_template = (command, scene_template)

    def predict(self, image_uri):
        with tempfile.TemporaryDirectory() as temp_dir:
            # Replace the scene  with the input scene.
            labels_path = os.path.join(temp_dir, "predictions.json")
            scene_config = make_scene_config(self.scene_template,
                                             self.scene_template.raster_source.raster_transformer.stats_uri,
                                             labels_path,
                                             [image_uri])
            app.logger.info(json_format.MessageToJson(scene_config))
            scene = scene_builder.build(scene_config, self.command.ml_task.get_class_map())
            self.command.scenes = [scene]

            # Run the predict command
            self.command.run()

            # Read in the result
            return json.loads(open(labels_path).read())

class CLModel(Model):

    def __init__(self):
        with tempfile.TemporaryDirectory() as temp_dir:
            super().__init__(PREDICT_PACKAGE_CL, temp_dir)

            # Initialize the model and set on the command
            model_path = download_if_needed(self.command.config.options.model_uri, temp_dir)
            model = model_builder.build_from_path(model_path)
            model._make_predict_function() # https://github.com/keras-team/keras/issues/2397#issuecomment-317255770
            self.command.ml_task.backend.model = model

class  ODModel(Model):
    def __init__(self):
        with tempfile.TemporaryDirectory() as temp_dir:
            super().__init__(PREDICT_PACKAGE_OD, temp_dir)

            # Initialize the model and set on the command
            model_path = download_if_needed(self.command.config.options.model_uri, temp_dir)
            detection_graph = load_frozen_graph(model_path)
            detection_graph.finalize()
            self.command.ml_task.backend.detection_graph = detection_graph
            self.command.ml_task.backend.session = tf.Session(graph=detection_graph)
```

This would ideally become a one- or few-liner with this new functionality.

lossyrob:
# API Refactor

This PR will merge in the feature branch for the API refactor. 

Supersedes #348 
Supersedes #367

Fixes #358
Fixes #362
Fixes #256
Fixes #237

## Goals of these changes

These changes were made to satisfy a number of goals:

### Introduce Builder pattern

There will be fluent builders for the following configuration types:

-Task
-Backend
-Dataset
--Train,Val,Predict (no builders)
--Scene
---RasterSource
----RasterTransformer
---LabelSource
---LabelStore
--Augmentation
-Analyzer
-Experiment

Commands:
-Analyze
--AnalyzeOptions
-Chip
--ChipOptions
-Train
--TrainOptions
-Predict
--PredictOptions
-Eval
--EvalOptions

The builders we use are in a fluent builder pattern.

## Backend config flexibility

Backend configs are stored as a Struct in protobuf, from a template.
The user can specify json configuration that overrides configuration
in an easy way - raster vision will search the json representation of
the config for the key, and replace it with the user supplied value.
If there are multiple keys in the config, an error will be thrown.

## Command configs

Command configuration will also have fluent builders. 

## Serialized Experiment as manifest

A serialized experiment describes completely the files and configuration needed to run each command. Each command has it's serialized config in it's output.

## Experiments can set configuration implicitly

There is a mechanism to set configuration implicitly - for example, a StatsAnalyzer does not need its URI set, but gets set based on experiment configuration, and the StatsTransformer has it's input URI set to the same URI.

### Flatten keys to use branch off workflows

-rv_root/
--analyze/default/
--chip/default/
--train/default/
--predict/default/
--eval/default/
--experiment/experiment-id.json

## Label Store tweaks

We have the same type of labels for scenes, and the label store differs by task type.
This is currently tangled in the system.
We should have label stores be based on labels lazily and based on configuration.

There's a difference between labels, which can either be raster or vector,
and the training labels that are required for training.
Labels + Task = TrainingLabels

Saving/Loading predictions is almost an entirely different thing. Simply needs to write and read the raster
or feature data.

Created LabelSource and LabelStore to differentiate.

## Customizable backend defaults

Allow configuraiton-based defaults for backends based on model key.
Allow this configuration to be overridden by users, so that they can
define their own model defaults for easy reuse.

## Easy unittest-like API for running experiments

For the upcoming release, we will only allow the `python script.py` style of running. In the future we will have discovery of experiments in a codebases for listing and running experiments.

### Running from JSON serialized commands

Commands are always run from JSON configuration. There will be configuration that will
allow plugins to be loaded from client code.

## Add Augmentation capability

An augmentor will work much like a RasterTransformer - it will manipulate the data being
generated from the RasterSource. The difference is, this will only happen for the training
data generation (`chip`), and can produce more data than the read data. Also Augmentors have to manipulate the labels along with the raster data.

## Plug architecture that allows users to define their own types

Includes:
-tasks
-backends
-analyzers
-augmentors
-raster_transformers
-evaltuators
-label sources
-label stores
-raster sources
-filesystem

## import rastervision as rv

We make all general use API points accessable
through the root package. This will allow us to simply `import rastervision as rv` and call everying off
of `rv._`, e.g. `rv.ExperimentConfig`

## Allow running only commands that are necessary in a set of experiments
When re-running experiments that have results from previous runs, don't re-run if results are already there. The exception is for training, when there are checkpoint files, you might want to "--continue-train". Also allow for re-running with "clobber" to overwrite past results.

## Allow running from a repository, that holds plugins across experiments and allows repeatability

The ideal scenario for repeatability is that we have a GitHub repository for each ML project.
That project uses the raster vision docker as is  or as a base.
Each command run is tagged with the Experiment it came from, the git commit hash for the repo it's
contained in, and the raster vision docker commit hash for the version of raster vision you are running.
Allowing repositories to be the main place for running raster vision will also allow plugins to be
specified on a per-project bases.

The plugin repository could be  a separate repository, so that users can maintain an organization-level  plugin repository.

## Run with other dependencies, without having to change out docker container

Allow hooks to run pip install on alternate dependencies, as well as running pip install on the target repository requirements.txt, so that other libraries could be installed. Perhaps a bootstrap.sh script can be used to run on every container run. This is a step below creating your own docker container based on RV and then deploying in Batch, etc.

# Analyzer

An Analyzer will take all the raster sources, download them, and run a processing step on them. You can have multiple analyzers for an experiment. Q: What happens when my raster data is too big for an Analyzer running on one machine? A: Use GeoTrellis for your preprocessing.

## Renames

Classification -> Chip Classification
MLTask -> Task
MLBackend -> Backend
LabelStore -> (LableSource, LabelStore)

## Package reorg

To make the code easier to navigate, items in core are broken out to their related packages,
along with configuration related to the package. Because packages contain multiple types and
not just implementations of a single type, the convention is to name the packages
as singular words. Therefore `ml_tasks` becomes `task`, `ml_backends` becomes `backend`, etc.

`core` is still used to contain core types, like `Box` and `Config`, that are not concepts that
have their own packages.

## Dry run flag

Add a dry run flag that prints out information about what commands will be run but does not run them.

## Unit Test reorg

Moved unit tests into a `tests` directory at the same level as the root `rastervision` package,
and renamed everyting to `test_*.py`. This is in line with what as far as I can tell is the standard,
allows for using defaults when running `unittest`, and makes excluding unit tests from distribution
packaging much easier.

To run unit tests, get into the docker container and run

```
python -m unittest discover tests
```

## Flexibility in backend installations

Users should be able to pip install rastervision without needing any backend library installed.
The imports that happen for backends should be done on-demand as backends are used.
That way you could pip install rastervision locally and utilize other functionality besides
training or running a model, or do so only with the target backend installed.

## Predicter

Users can use prediction-packages to run the Predictor, either from code or from the command line. The Predictor can load a model in case it will be used for loading the model once and doing many predictions. Fixes #368


lewfish:
Also, `Default*Provider` should be changed to `*DefaultProvider`

lewfish:
I said `TrainingLabelSource` above but `GroundTruthLabelSource` makes more sense since it is also used in evaluation.

src/rastervision/backend/tf_object_detection_config.py
```diff
@@ -0,0 +1,352 @@
...
+           TF object detection pipeline config as necessary.
+        """
+        self.validate()
+
+        b = deepcopy(self)
+
+        # Check if a pretrained model was assigned.
+        pretrained_model = b.config.get('pretrained_model_uri')
+        if pretrained_model:
+            b = b.with_config({'fineTuneCheckpoint': pretrained_model})
```
lewfish: I think that setting this has no effect since the backend overwrites `fine_tune_checkpoint` to the local path of the downloaded pretrained model here https://github.com/azavea/raster-vision/blob/feature/api-refactor/src/rastervision/backend/tf_object_detection.py#L495

But even if that wasn't occurring, this wouldn't work because TFOD can't handle S3 URIs.

lewfish: Nevermind this comment. I see that this works now.

PR #384, opened 2018-09-15
Move  to Sphynx documentation for ReadTheDocs

lossyrob:
## Overview

Moves our current docs to restructured text and a sphynx setup, so we can use ReadTheDocs


lewfish:
What's the command to build the docs and push to CI? Is that going to be part of another PR?

lossyrob:
ReadTheDocs pics up the docs automatically, this doesn't happen through CI.

PR #414, opened 2018-09-25
Remove deployment code from this repository

### LINKED ISSUE #245: Make `raster-vision-ops` Docker image
We would like to remove Vagrant and Ansible from this project and just use Docker. However, our Terraform and Packer stuff is still in that setup. We need to make a `raster-vision-ops` image (or something like that) which will let us perform these functions. This is a low priority.

lossyrob:
Deployment code is now hosted at https://github.com/azavea/raster-vision-aws

Fixes #245 

Depends on #409


lossyrob:
@lewfish removed those files. I removed references to Vagrant et al in the  README and docs - updating the documentation can be handled in another PR.

That commit is hanging around, but was part of another PR and is already in the codebase, so ignore that commit.

PR #425, opened 2018-09-26
Implement `--dry-run` Flag

### LINKED ISSUE #403: Implement `--dry-run` / `-n` option in CLI
We should have a `--dry-run`/`-n` option in the CLI which will print information about the commands to be run, but not actually run them. The CLI option exists, but needs to be implemented in the code.

jamesmcclain:
## Overview

### Checklist

- [ ] Modify remote runner

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

Closes #403


lossyrob:
The dry run option seems pretty pervasive in the command  configs,  runners, etc. I wonder if there is a less  invasive set of changes that gets to the same results, but uses a top level runner to do the work? i.e. create a `DryRunRunner` that "runs" the experiment by describing the configuration for commands it would run if it were a real runner? And then the skipped commands could be handled by the top level ExperimentRunner, or else delegated to inherited members, in which case the DryRunRunner would report on the skipped commands. What do you think?

lossyrob:
OK, happy to discuss further if reverse course isn't ideal. 

One other point I thought about is, because dry-run is (was) implemented based on `create_command`, which doesn't get called on the 'calling' machine when running remotely, so that output would end up in a cloudwatch log for the aws_runner - if we base the dry run purely on the command configs and what will be/won't be run (maybe with a verbose option of even dumping the command config's proto), then a local dry run would describe work that would be done against AWS

src/rastervision/command/predict_command.py
```diff
@@ -12,3 +12,14 @@ def run(self, tmp_dir):
         msg = 'Making predictions...'
         click.echo(click.style(msg, fg='green'))
         self.task.predict(self.scenes, tmp_dir)
+
+
+class NoOpPredictCommand(NoOpCommand):
```
lossyrob: This seems leftover from the previous changes (along with predict_command_config changes)

jamesmcclain:
Changes made.

PR #408, opened 2018-09-20
Refactor semantic segmentation

lewfish:
## Overview

This PR adds back the segmentation code in a refactored form to be compatible with the new API. It attempts to maintain the original behavior except for making each prediction GeoTIFF file match the georeferencing and extent of the corresponding input GeoTIFF. Other improvements for segmentation are left to other issues including: #417 #418 #420 #429 #430

## Testing Instructions

* You can test this on Potsdam using this [experiment file](https://github.com/azavea/raster-vision/files/2429795/potsdam.py.txt) You'll need to modify the `root_uri` and `data_uri`. 
* If you're running things out of the RV Docker container, place the file in `src/rastervision`
* To run locally use something like 
```
python -m rastervision run local \
    -e rastervision.potsdam \
    -a tmp_dir /opt/data/tmp  -a local_run True
```
* To run remotely, use this .rastervision config
```
[AWS_BATCH]
job_queue=raster-vision-gpu
job_definition=raster-vision-gpu-lewfish
```
and then run something like 
```
python -m rastervision run aws_batch \
    -e rastervision.potsdam \
    -a local_run False
```

After training for 100k steps with Mobilenet, we get the following predictions and eval which is roughly in line with previous results. We could probably get better results using a bigger model and more steps.
screen shot 2018-09-28 at 3 20 54 pm

```
[
    {
        "precision": 0.8656729563616176,
        "gt_count": 1746655,
        "class_id": 1,
        "recall": 0.8081258176342782,
        "count_error": 200350.09143477102,
        "class_name": "Car",
        "f1": 0.8351868892794376
    },
    {
        "precision": 0.9077043151132905,
        "gt_count": 28166583,
        "class_id": 2,
        "recall": 0.9453450210840271,
        "count_error": 1496358.1113330645,
        "class_name": "Building",
        "f1": 0.9259374145605163
    },
    {
        "precision": 0.8105826727015737,
        "gt_count": 30140893,
        "class_id": 3,
        "recall": 0.8826813459043832,
        "count_error": 3813131.239710051,
        "class_name": "Low Vegetation",
        "f1": 0.8448803483993653
    },
    {
        "precision": 0.8853166963497794,
        "gt_count": 16928529,
        "class_id": 4,
        "recall": 0.7333917790494379,
        "count_error": 2298428.025324646,
        "class_name": "Tree",
        "f1": 0.798672495115001
    },
    {
        "precision": 0.8905422564785969,
        "gt_count": 29352493,
        "class_id": 5,
[...]

PR #431, opened 2018-09-27
Make validation chips

lewfish:
## Overview

This PR fixes a typo that made it so the validation chips were the same as the training chips.

## Testing Instructions

* Run the chip command for an experiment where the training and validation scenes differ and observe the output. I did this for the Potsdam experiment in #408


lewfish:
Superceded by #432

PR #441, opened 2018-10-02
Temporary Directory Machinery

### LINKED ISSUE #400: Rework/Clean up the way we set the temporary directory to use
This block of code needs to be refactored into a proper place, and cleaned up a bit  to make it more clear exactly what it does. The temporary directory configuration should be a setting inside the raster vision configuration (rv_config) and also a command line argument (in the `main` method in `cli/main.py`

https://github.com/azavea/raster-vision/blob/develop/src/rastervision/utils/files.py#L298-L326

jamesmcclain:
## Overview

- [x] This block of code needs to be refactored into a proper place, and cleaned up a bit to make it more clear exactly what it does.
- [x] The temporary directory configuration should be a setting inside the raster vision configuration (`rv_config`) and also a command line argument (in the `main` method in `cli/main.py`).
- ~~I noticed in the refactor that you made it so `tmp_dir` is passed into a lot of functions. I think this is unnecessary because this line of code makes it so that all temp dirs created with `tempfile` are inside `RV_TEMP_DIR`.~~  (Will be handled in subsequent pull request.)
- ~~Cleanup with `atexit`~~
- [x] Avoid use of `tempfile.tempdir` as global variable

### Checklist

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

Closes #400

### DIFF
--- a/src/rastervision/rv_config.py
+++ b/src/rastervision/rv_config.py
@@ -1,6 +1,9 @@
 import os
 import json
 import tempfile
+import shutil
+import atexit
+
 from pathlib import Path
 
 from everett.manager import (ConfigManager, ConfigDictEnv, ConfigEnvFileEnv,
@@ -27,7 +30,7 @@ class RVConfig:
         return tempfile.TemporaryDirectory()
 
     @staticmethod
-    def set_tmp_dir(tmp_dir=None, delete_tmp_dir_atexit=False):
+    def set_tmp_dir(tmp_dir=None, rm_tmp_dir=False):
         """Set RVConfig.tmp_dir to well-known value.
 
         This static method sets the value of RVConfig.tmp_dir to some
@@ -75,6 +78,11 @@ class RVConfig:
 
         finally:
             make_dir(RVConfig.tmp_dir)
+            tmp_dir = RVConfig.tmp_dir
+            if rm_tmp_dir:
+                atexit.register(lambda: shutil.rmtree(tmp_dir))
+                print('{} directory will be removed upon nominal exit.'.format(
+                    tmp_dir))
             print('Temporary directory is: {}'.format(RVConfig.tmp_dir))
 
     @staticmethod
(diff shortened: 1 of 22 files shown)


lossyrob:
I like this approach. I'm wondering if there's some way to address the comment

> I noticed in the refactor that you made it so tmp_dir is passed into a lot of functions. I think this is unnecessary because this line of code makes it so that all temp dirs created with tempfile are inside RV_TEMP_DIR.

with utilizing `atexit` to clean up some global temporary directory.

The dependency injection that's being done for passing around temporary directories might have some aspects that would be rolled back by moving to a global temp dir. I'm going to take a look at the code to see if I can find anything that would cause this approach to be undesirable.

jamesmcclain:
Because directories created by `tempdir.TemporaryDirectory()` are automatically removed, I assumed that the directory to be deleted upon exit is the top-level temporary directory.  That is obviously dangerous, so that only happens when the `--rm-tmp-dir` flag is given on the command line.

src/rastervision/rv_config.py
```diff
@@ -1,21 +1,95 @@
...
+    # it will be mirrored on the host file system which is needed for
+    # running in a Docker container with limited space on EC2.
+    DEFAULT_DIR = '/opt/data/tmp/'
+
+    tmp_dir = None
+
+    @staticmethod
+    def get_tmp_dir():
+        tempfile.tempdir = RVConfig.tmp_dir
+        return tempfile.TemporaryDirectory()
```
lossyrob: @jamesmcclain This sets the global tempdir. Can you use:

```
tempfile.TemporaryDirectory(dir=RVConfig.tmp_dir)
```
instead?

jamesmcclain: Ah, okay, much appreciated.

lossyrob:
Merged these changes into #446

PR #451, opened 2018-10-05
Unthread `tmp_dir`

### LINKED ISSUE #447: Remove Unnecessary Use of `tmp_dir`
https://github.com/azavea/raster-vision/issues/400#issuecomment-423300643

See #400 
See #441

jamesmcclain:
## Overview

Attempt to remove detect and correct all instances where `tmp_dir` (or similar) is (was) being unnecessarily passed.  (Unnecessary in light of the fact that the location of the temporary directory is available globally.)

I have *not* removed the `tmp_dir` (or similar) parameter, only made it optional and not used it where possible.

I acknowledge that this is essentially a negative result, but I made several attempts toward complete removal and in each case it looked as though it was going to be a sizable change -- deceptively large.

### Checklist

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

Closes #447

### DIFF
--- a/src/rastervision/command/command.py
+++ b/src/rastervision/command/command.py
@@ -7,6 +7,12 @@ class Command(ABC):
         """Run the command."""
         pass
 
+    def set_tmp_dir(self, tmp_dir):
+        self._tmp_dir = tmp_dir
+
+    def get_tmp_dir(self):
+        return self._tmp_dir
+
 
 class NoOpCommand(Command):
     """Defines a command that does nothing.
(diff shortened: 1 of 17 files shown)


src/rastervision/command/command.py
```diff
@@ -7,6 +7,15 @@ def run(self, tmp_dir):
         """Run the command."""
         pass
 
+    def set_tmp_dir(self, tmp_dir):
+        self._tmp_dir = tmp_dir
+
+    def get_tmp_dir(self):
```
lossyrob: Should this check for `None` or a missing property, and then call `RVConfig.get_tmp_dir()`? If someone creates a command not through a config, and doesn't call `set_tmp_dir`, I'm not sure it should break.

PR #457, opened 2018-10-05
New readme

### LINKED ISSUE #442: Update repo README
Streamline description, add screenshots, add guiding principles, easy install, example experiment, and resources.

lewfish:
## Overview

This is a start on the new readme. I think further work will require discussion. At first I was going to add your typical install and quickstart instructions when I realized that that info is in the raster-vision-examples readme, and that we should just point people there. So instead, I'd like to use the RV repo readme to show an illustrative example (real, but lightly edited) of an experiment configuration and the command line invocation to run it. The idea is to just give people an impression of what the library does and whet their appetite for reading more detailed documentation elsewhere. 

Closes #442


lossyrob:
I think the approach is sound. Seems like maybe we want to place the lists of things (e.g. generate training chips, train a model, make predictions, etc) into bullet point lists to make them more easy on a quick read through, and add the images of each task that you mentioned.

PR #458, opened 2018-10-05
Add logging

### LINKED ISSUE #260: Use logger instead of print statements

nholeman:
## Overview

This PR replaces print statements (and sometimes calls to
click.echo) with Python logs.

### Checklist

### Notes

Logging level can be set with the CLI verbosity option.

If a log is at or above the logging level, the log will
go to stdout.

Closes #260

### DIFF
--- a/src/rastervision/cli/main.py
+++ b/src/rastervision/cli/main.py
@@ -3,15 +3,18 @@ import sys
 import os
 
 import click
+import logging
 
 import rastervision as rv
 from rastervision.experiment import (ExperimentLoader, LoaderError)
 from rastervision.runner import (ExperimentRunner)
 from rastervision.rv_config import RVConfig
 
+log = logging.getLogger(__name__)
+
 
 def print_error(msg):
-    click.echo(click.style(msg, fg='red'), err=True)
+    log.error(msg)
 
 
 @click.group()
--- a/src/rastervision/runner/experiment_runner.py
+++ b/src/rastervision/runner/experiment_runner.py
@@ -1,37 +1,38 @@
 from abc import (ABC, abstractmethod)
 from typing import (List, Union)
+import logging
 
-import click
 from google.protobuf import json_format
 
 import rastervision as rv
 from rastervision.runner import (CommandDefinition, CommandDAG)
 from rastervision.cli import Verbosity
 
+log = logging.getLogger(__name__)
+
 
 class ExperimentRunner(ABC):
-    def print_command(self, command_def, command_id=None, command_dag=None):
+    def log_command(self, command_def, command_id=None, command_dag=None):
         verbosity = Verbosity.get()
         command_type = command_def.command_config.command_type
         experiment_id = command_def.experiment_id
-        click.echo(
-            click.style('{} '.format(command_type), bold=True), nl=False)
-        click.echo('from {}'.format(experiment_id))
+        log.info('{} '.format(command_type))
+        log.info('from {}\n'.format(experiment_id))
 
         if verbosity >= Verbosity.VERBOSE:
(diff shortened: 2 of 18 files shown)


src/rastervision/cli/main.py
```diff
@@ -3,15 +3,18 @@
...
 
 import click
+import logging
 
 import rastervision as rv
 from rastervision.experiment import (ExperimentLoader, LoaderError)
 from rastervision.runner import (ExperimentRunner)
 from rastervision.rv_config import RVConfig
 
+log = logging.getLogger(__name__)
```
lossyrob: I think that since the CLI is explicitly  for the command line, we'd want to log errors to the console with the `click.echo` method we have, and not through the logger.

src/rastervision/runner/experiment_runner.py
```diff
@@ -1,37 +1,38 @@
...
+import logging
 
-import click
 from google.protobuf import json_format
 
 import rastervision as rv
 from rastervision.runner import (CommandDefinition, CommandDAG)
 from rastervision.cli import Verbosity
 
+log = logging.getLogger(__name__)
```
lossyrob: Experiment Runner output is similar to the CLI output - the output is meant for the console, and the coloring is important. This should be moved back to the echo commands.

src/rastervision/rv_config.py
```diff
@@ -94,6 +97,12 @@ def __init__(self,
                  verbosity=Verbosity.NORMAL):
         self.verbosity = verbosity
 
+        # Set logging level
+        log = logging.getLogger('rastervision')
+        if self.verbosity >= Verbosity.NORMAL:
+            log.setLevel(logging.INFO)
```
lossyrob: Logging in the "very verbose" case should be set to `DEBUG`, otherwise INFO

src/rastervision/rv_config.py
```diff
@@ -94,6 +97,12 @@ def __init__(self,
...
+        log = logging.getLogger(__name__)
```
lossyrob: Is `log` unused here? If this is to avoid setting a logger that is based on `rastervision` since it's used above, I'd recommend giving that variable another name (e.g. `root_log = logging.getLogger('rastervision')`)

src/rastervision/task/task.py
```diff
@@ -171,7 +169,6 @@ def predict_batch(predict_chips, predict_windows):
             new_labels = self.backend.predict(
                 np.array(predict_chips), predict_windows, tmp_dir)
             labels += new_labels
-            print('.' * len(predict_chips), end='', flush=True)
```
lossyrob: We shouldn't get rid of  this, it's useful to have this output. I think it's fine to leave this as a print if we can't accommodate the `end=''` in the logger.

src/rastervision/rv_config.py
```diff
@@ -74,13 +77,13 @@ def set_tmp_dir(tmp_dir=None):
...
         except Exception as e:
-            print(
+            log.warning(
                 'Root temporary directory cannot be used: {}. Using root: {}'.
                 format(explicit_tmp_dir, RVConfig.DEFAULT_DIR))
             RVConfig.tmp_dir = RVConfig.DEFAULT_DIR
         finally:
             make_dir(RVConfig.tmp_dir)
-            print('Temporary directory is: {}'.format(RVConfig.tmp_dir))
+            log.info('Temporary directory is: {}'.format(RVConfig.tmp_dir))
```
lossyrob: Can you make this a `log.debug`, so that it doesn't display unless you ask for verbose?

lossyrob:
I updated these changes in #466.

PR #495, opened 2018-10-15
note on tensorflow versions

simonkassel:
## Overview

Currently due to a TensorFlow issue, the quickstart failed on an older computer. This PR adds a note about this issue in the quickstart docs.


simonkassel:
I did my best to leave a useful note given my understanding of the problem but let me know if there is something I should add.

PR #499, opened 2018-10-15
Document macOS setup instructions

### LINKED ISSUE #487: h5py failure when running `pip install` on macOS
I tried out running `pip install` on two different Macbook Pros each running High Sierra and encountered a failure that seems to be related to the installation of `h5py`.

The ultimate failure message is:

```
Command "/usr/local/opt/python/bin/python3.7 -u -c "import setuptools, tokenize;__file__='/private/var/folders/7r/g5z55rkd0hlcv_h8bbv1rbx00000gn/T/pip-install-c_k7qlul/pyproj/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /private/var/folders/7r/g5z55rkd0hlcv_h8bbv1rbx00000gn/T/pip-record-_w6_lqru/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /private/var/folders/7r/g5z55rkd0hlcv_h8bbv1rbx00000gn/T/pip-install-c_k7qlul/pyproj/   
```

Attaching the full log as a text file.
[rastervision_install.txt](https://github.com/azavea/raster-vision/files/2469813/rastervision_install.txt)

kellyi:
## Overview

This PR updates the Raster Vision docs to include macOS specific setup instructions which I used to circumvent the problems I encountered in #487.

The minimal set of manually installation requirements seems to be:

```
pip install cython
pip install git+https://github.com/jswhit/pyproj.git
```

...before `pip install rastervision`. The git+https install for pyproj seems to be required on macOS, according to https://github.com/jswhit/pyproj/issues/138 and https://github.com/jswhit/pyproj/issues/136 

## Testing Instructions

From a clean install on macOS High Sierra, follow the the new macOS specific setup instructions and verify that they work.

After installation, you should be able to run

```
$ python3
>>> import rastervision
```

...without error.

Closes #487


docs/setup.rst
```diff
@@ -16,6 +16,20 @@ You can get the library directly from PyPI:
 
 .. note:: Raster Vision requires Python 3 or later.
 
+macOS Installation
```
lossyrob: I did not have to do this for my MacOS installation, for whatever reason - it could be because the necessary libraries were installed with other dependencies previously. Considering that, should we frame this more as a "if you have the following issues, you might need to do this" rather than as the definitive Mac install instructions? That way we don't encourage folk to install an unreleased version of pyproj if they don't have to.

I think another thing besides Mac that's worth maybe mentioning is that it's a python 3.7 issue (which was mentioned in the pyproj  issues you linked to in #487)

kellyi: Ah, okay -- that makes sense. If we're okay keeping the header, I can change it to something like...

##### Troubleshooting macOS Installation

If you encounter problems running `pip install rastervision` on macOS, you may have to manually install Cython and pyproj. To circumvent an issue installing pyproj with Python 3.7, you may also have to install it using `git+https`:

... instructions

##### Tensorflow header

....

lossyrob: I think that makes sense. Thanks!

kellyi: Cool. Updated and squashed!

lossyrob:
Great, thank you. I'll merge once Travis goes green.

lewfish:
I cherry picked the commit to the 0.8 branch so it should be on the docs site soon.

PR #538, opened 2018-10-26
Optimize update_for_command

lossyrob:
## Overview

This PR refactors the update_for_command architecture to optimize the code - when using Raster Vision for experiments of lots of files (e.g. 250,000), the update_for_command chain would take way too long. It also replaces the deepcopy mechanism for configs to use pickle, which is faster than deepcopy.

Other changes:
- Use `scandir` instead of listdir when appropriate to avoid iterating over directorys with many files.
- Added debugging output for experiment running, which can be triggered with the verbose flag (`-v`)
- Don't deepcopy Datasets, just make a  copy  of the list

### Notes

Moving the config  __deepcopy__ mechanism to use pickle has the side effect of requiring that configs hold only serializable items. I've run into places  where in  `from_proto` we don't wrap a `repeated` item in a list, and that was causing serialization errors. Make sure to hold only lists and other serializable items and not raw protobuf repeated field message type.

### DIFF
--- a/rastervision/data/scene_config.py
+++ b/rastervision/data/scene_config.py
@@ -107,39 +107,31 @@ class SceneConfig(BundledConfigMixin, Config):
     def to_builder(self):
         return SceneConfigBuilder(self)
 
-    def update_for_command(self, command_type, experiment_config,
-                           context=None):
+    def update_for_command(self,
+                           command_type,
+                           experiment_config,
+                           context=None,
+                           io_def=None):
         if context is None:
             context = []
         context = context + [self]
-        io_def = rv.core.CommandIODefinition()
+        io_def = io_def or rv.core.CommandIODefinition()
 
-        b = self.to_builder()
-
-        (new_raster_source,
-         sub_io_def) = self.raster_source.update_for_command(
-             command_type, experiment_config, context)
-        io_def.merge(sub_io_def)
-        b = b.with_raster_source(new_raster_source)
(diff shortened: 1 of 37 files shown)


rastervision/data/scene_config.py
```diff
@@ -107,39 +107,31 @@ def create_local(self, tmp_dir):
...
+        self.raster_source.update_for_command(command_type, experiment_config,
+                                              context, io_def)
 
         if self.label_source:
-            (new_label_source,
-             sub_io_def) = self.label_source.update_for_command(
-                 command_type, experiment_config, context)
-            io_def.merge(sub_io_def)
-            b = b.with_label_source(new_label_source)
+            self.label_source.update_for_command(
```
lewfish: Don't we need to capture the `io_def` returned by this call and then merge with `io_def`?

lossyrob: No, because now with the mutable io_def, the are just passed forward and configs all sequentially modify the same io_def instance. So when a label_source adds to the io_def, that's the same io_def that is further modified by the label_store etc

lewfish: Oh, right. So I guess that means we don't need to return io_def then.

lewfish: Actually that's not quite true because io_def defaults to None in the parameter list.

lewfish:
Ok to merge?

PR #539, opened 2018-10-26
Implement activation to avoid keeping files open.

lossyrob:
## Overview

This  PR implements  an 'activation' architecture that allows for scene components to be opened and closed as needed. This  change arises from seeing "Too many files open" errors when dealing with a large number of files - the way the code works currently is that files (like the GeoTIFFs of a raster source) remain open for the duration of the lifetime of the entity. This uses  context managers to allow  code control over when things are opened and closed.

### Notes

One thing I was hoping would work out is to put off downloading until the raster source was activated. We need information from the raster up front (like it's crs transform and how many channels), and trying to do that through VSI functionality seemed like a secondary change.

### DIFF
--- a/rastervision/core/raster_stats.py
+++ b/rastervision/core/raster_stats.py
@@ -17,20 +17,18 @@ class RasterStats():
 
         def chip_stream(channel):
             for raster_source in raster_sources:
-                windows = raster_source.get_extent().get_windows(
-                    chip_size, stride)
-                for window in windows:
-                    chip = raster_source.get_raw_chip(window).astype(
-                        np.float32)
-                    chip = chip[:, :, channel].ravel()
-                    # Ignore NODATA values.
-                    chip[chip == 0.0] = np.nan
-                    yield chip
+                with raster_source.activate():
+                    windows = raster_source.get_extent().get_windows(
+                        chip_size, stride)
+                    for window in windows:
+                        chip = raster_source.get_raw_chip(window).astype(
+                            np.float32)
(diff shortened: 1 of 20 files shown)


rastervision/core/raster_stats.py
```diff
@@ -17,20 +17,18 @@ def compute(self, raster_sources):
...
+                        chip_size, stride)
+                    for window in windows:
+                        chip = raster_source.get_raw_chip(window).astype(
+                            np.float32)
+                        chip = chip[:, :, channel].ravel()
+                        # Ignore NODATA values.
+                        chip[chip == 0.0] = np.nan
+                        yield chip
+
+        nb_channels = len(raster_sources[0].channel_order)
```
lewfish: I think we actually want to compute the stats for all channels, not just the ones selected by channel_order. Things downstream could break because of this change?

rastervision/data/activate_mixin.py
```diff
@@ -0,0 +1,98 @@
...
+            self.deactivate()
+
+        @classmethod
+        def dummy(cls):
+            def noop():
+                pass
+
+            return cls(noop, noop)
+
+    class CompositContextManager:
```
lewfish: typo: Composite

rastervision/core/raster_stats.py
lossyrob: Ah sorry I must have got confused by which was approved - I'll address these comments as a follow up.

Issue #508, opened 2018-10-17
Periodic syncing not working for tf_deeplab


lewfish:
I've been running a job for 6 hours and the only checkpoints on S3 are from around 6 hours ago. This is despite the fact that the job was configured to sync ever 600 seconds.

PR #555, opened 2018-11-07
Wrap `sync_to_dir` in Loop

jamesmcclain:
## Overview

This gives `start_sync` the correct behavior.

Note that there is still a `threading.Timer` in there instead of a `threading.Thread`.  The reason is that the latter does not have a `__cancel__` method which prevents it from (I believe) interacting well with the `with` keyword.

Note that the previous implementation seems to have never worked.  Please see the discussion [here](https://stackoverflow.com/questions/3393612/run-certain-code-every-n-seconds).  This can be further proven by typing the following into a python shell:

## Further Proof ##

Play with variations of this code in a Python interpreter:

```python
import threading

def hello():
     print("hello")

t = threading.Timer(1, hello)
t.daemon = True
t.start()
```

with the number of seconds given to `Timer` either `1` or `1000` and with the `t.daemon = True` line either absent or present.  Give a ctrl-D to the REPL 10 seconds after `t.start()` and observe the results. 
 The collective results will imply that `Timer` runs the code once and only once and there is no thread that persists after the first run.

### Checklist

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

Closes https://github.com/azavea/raster-vision/issues/508

### DIFF
--- a/rastervision/utils/files.py
+++ b/rastervision/utils/files.py
@@ -1,9 +1,8 @@
 import os
 import shutil
+from threading import Timer
 import time
 import logging
-
-from threading import Timer
 
 from google.protobuf import json_format
 
--- a/rastervision/utils/files.py
+++ b/rastervision/utils/files.py
@@ -1,7 +1,9 @@
 import os
 import shutil
-from threading import Timer
+import time
 import logging
+
+from threading import Timer
 
 from google.protobuf import json_format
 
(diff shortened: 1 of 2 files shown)


rastervision/utils/files.py
```diff
@@ -1,8 +1,10 @@
 import os
 import shutil
-from threading import Timer
+import time
 import logging
 
+from threading import Timer
```
lossyrob: nitpick - We've been keeping the structure as

```python
[library imports]

[3rd party imports]

[rastervision imports]
```

Is there a reason to add the extra section? If not we should keep to that convention, as a tidiness thing.

jamesmcclain: Ah, I see.  I had not recognized the pattern.

PR #556, opened 2018-11-07
changes to configbuilders

### LINKED ISSUE #537: Use 'self' in ConfigBuilder 'from_proto' instead  of constructing new builder
Example: https://github.com/azavea/raster-vision/blob/f6ea64a37fd4d09375da1838cd679e6cbce5b35b/rastervision/task/semantic_segmentation_config.py#L91

We  can just use `self` here instead of instantiating new objects; that way subclasses (e.g. plugins) can utilize the `from_proto` methods of  their parent classes.

We should check for other places this is happening.

simonkassel:
## Overview

This pr changes the `from_proto` method of a number of configbuilders to use self rather than instantiating new configbuilder objects

### Checklist

## Testing Instructions

* Run integration tests

Closes #537


simonkassel:
@lossyrob is this ready to merge?

PR #562, opened 2018-11-13
check value types in validate methods

### LINKED ISSUE #519: `validate` methods should check types
Each ConfigBuilder class has a `validate` method that checks that various fields have been set. These should also check that the values are of the correct type. I forgot to call `build()` on a backend, and when I built the experiment (which depends on the backend), I got the following error which could be confusing to a new user.
```
File "/opt/src/rastervision/experiment/experiment_config.py", line 54, in update_for_command
  new_backend, sub_io_def = self.backend.update_for_command(
AttributeError: 'TFObjectDetectionConfigBuilder' object has no attribute 'update_for_command'
```
If we checked in `ExperimentConfigBuilder.validate` that the backend was of type `BackendConfig`, this wouldn't have happened.

simonkassel:
## Overview

this PR updates the validate methods in configBuilders to check whether values are the correct types

### Checklist

## Testing Instructions

* run integration tests

Closes #519

### DIFF
--- a/rastervision/command/chip_command_config.py
+++ b/rastervision/command/chip_command_config.py
@@ -88,7 +88,8 @@ class ChipCommandConfigBuilder(CommandConfigBuilder):
                 'with_train_scenes or with_experiment')
         if len(self.train_scenes) > 0:
             if not isinstance(self.train_scenes[0], SceneConfig):
-                raise rv.ConfigError('train_scenes must be a list of class SceneConfig, got a list of {}'.
+                raise rv.ConfigError(
+                    'train_scenes must be a list of class SceneConfig, got a list of {}'.
                     format(type(self.train_scenes[0])))
         if self.val_scenes == []:
             raise rv.ConfigError(
@@ -96,9 +97,9 @@ class ChipCommandConfigBuilder(CommandConfigBuilder):
                 'with_val_scenes or with_experiment')
         if len(self.val_scenes) > 0:
             if not isinstance(self.val_scenes[0], SceneConfig):
-                raise rv.ConfigError('val_scenes must be a list of class SceneConfig, got a list of {}'.
+                raise rv.ConfigError(
+                    'val_scenes must be a list of class SceneConfig, got a list of {}'.
                     format(type(self.val_scenes[0])))
-
 
     def build(self):
         self.validate()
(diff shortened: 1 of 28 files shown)


rastervision/experiment/experiment_config.py
```diff
@@ -163,6 +164,10 @@ def __init__(self, prev=None):
 
     def validate(self):
 
+        backend = self.config.get('backend')
+        if not issubclass(type(backend), BackendConfig):
+            raise rv.ConfigError('Call "build()" on the backend.')
```
lossyrob: Not calling  `build` on a `BackendConfigBuilder` is one way for the types not to line up - but not all the ways the type won't match. This error should be more like:

`'Backend set with "with_backend" needs to be of type BackendConfig, got {}'.format(type(backend))`

simonkassel: yeah, that makes sense, thanks

rastervision/data/label_source/chip_classification_geojson_source_config.py
```diff
@@ -86,6 +86,7 @@ def validate(self):
             raise rv.ConfigError(
                 'You must set the uri for ChipClassificationGeoJSONSourceConfig'
                 ' Use "with_uri".')
+        check_uri_type(self.config.get('uri'))
```
jamesmcclain: There are some changes that seemed to have happened underneath this PR branch.  This line in particular changes somewhat (see [here](https://github.com/jamesmcclain/raster-vision/commit/780c84f91b1857d5637fd3e9d991c628f016fe97)).

rastervision/command/chip_command_config.py
```diff
@@ -75,17 +77,29 @@ def validate(self):
...
             raise rv.ConfigError('Backend not set for ChipCommandConfig. Use '
                                  'with_backend or with_experiment')
+        check_backend_type(self.backend)
         if self.train_scenes == []:
             raise rv.ConfigError(
                 'Train scenes not set for ChipCommandConfig. Use '
                 'with_train_scenes or with_experiment')
+        if len(self.train_scenes) > 0:
+            if not isinstance(self.train_scenes[0], SceneConfig):
+                raise rv.ConfigError(
```
lewfish: It would be preferable loop through the whole list and check each item, rather than only check the first one. (Same comment applies to some other spots with lists.)

simonkassel: yeah, that makes sense. I made these changes where necessary

simonkassel:
@lewfish is this good to merge?

lewfish:
> @lewfish is this good to merge?

+1

PR #575, opened 2018-11-26
Feature Extraction

### LINKED ISSUE #564: Improvements to feature extraction

### LINKED ISSUE #611: Evaluation of Vector Outputs
This was left as future work following https://github.com/azavea/raster-vision/pull/575 .

### LINKED ISSUE #612: "polygons" Mode Vector Output
The pull request https://github.com/azavea/raster-vision/pull/575 adds "`buildings`" mode, "`polygons`" mode was left as future work.

jamesmcclain:
## Overview

Adds ability to produce a GeoJSON file containing semantic segmentation predictions.

The ProtoBuf schema has been updated to include a [list of vector outputs](https://github.com/azavea/raster-vision/pull/575/files#diff-ba58680523016964b2532a3779c09392).

The segmentation integration test gives the following results:

mask
screenshot from 2018-12-05 14-38-44

Evaluation of vector outputs has been left for future work (see https://github.com/azavea/raster-vision/issues/611 ).

`polygon` output mode left for future work (see https://github.com/azavea/raster-vision/issues/612 ).

### Checklist

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

Closes https://github.com/azavea/raster-vision/issues/544
Closes #611
Closes #612
Closes #564

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,9 +11,9 @@ RUN add-apt-repository ppa:ubuntugis/ppa && \
     apt-get update && \
     apt-get install -y wget=1.* git=1:2.* python-protobuf=2.* python3-tk=3.* \
                        gdal-bin=2.1.* \
-                       jq \
-                       build-essential libsqlite3-dev zlib1g-dev \
-                       libopencv-dev python-opencv && \
+                       jq=1.5* \
+                       build-essential libsqlite3-dev=3.11.* zlib1g-dev=1:1.2.* \
+                       libopencv-dev=2.4.* python-opencv=2.4.* && \
     apt-get autoremove && apt-get autoclean && apt-get clean
 
 # Install protoc
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,3 +73,7 @@ RUN cat extras_requirements.json | jq  '.[][]' | grep -v 'tensorflow' | sort -u 
 # Install requirements.txt
 COPY ./requirements.txt /opt/src/requirements.txt
 RUN pip install -r requirements.txt
+
+# Install optional-requirements.txt
+COPY ./optional-requirements.txt /opt/src/optional-requirements.txt
+RUN pip install -r optional-requirements.txt
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,13 +6,14 @@ FROM tensorflow/tensorflow:1.10.1-gpu-py3
 ENV LC_ALL C.UTF-8
 ENV LANG C.UTF-8
 
-# Install misc. packages, GDAL, jq (for parsing extra-requirements.json), and Tippecanoe
+# Install misc. packages, GDAL, jq (for parsing extra-requirements.json), Tippecanoe dependencies, and OpenCV
 RUN add-apt-repository ppa:ubuntugis/ppa && \
     apt-get update && \
     apt-get install -y wget=1.* git=1:2.* python-protobuf=2.* python3-tk=3.* \
-            gdal-bin=2.1.* \
-            jq \
-            build-essential libsqlite3-dev zlib1g-dev && \
+                       gdal-bin=2.1.* \
+                       jq \
+                       build-essential libsqlite3-dev zlib1g-dev \
+                       libopencv-dev python-opencv && \
     apt-get autoremove && apt-get autoclean && apt-get clean
 
 # Install protoc
@@ -46,6 +47,7 @@ ENV CURL_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
 RUN rm -f /usr/bin/pip && ln -s /usr/bin/pip3 /usr/bin/pip
 RUN rm -f /usr/bin/python && ln -s /usr/bin/python3 /usr/bin/python
 
+# Install Tippecanoe
 RUN cd /tmp && \
     wget https://github.com/mapbox/tippecanoe/archive/1.32.5.zip && \
     unzip 1.32.5.zip && \
--- a/integration_tests/semantic_segmentation_tests/experiment.py
+++ b/integration_tests/semantic_segmentation_tests/experiment.py
@@ -55,7 +55,7 @@ class SemanticSegmentationIntegrationTest(rv.ExperimentSet):
                                            .with_raster_source(label_path) \
                                            .build()
 
-        vector_output = {'mode': 'buildings', 'class_id': 1}
+        vector_output = {'denoise': 1, 'mode': 'buildings', 'class_id': 1}
         label_store = rv.LabelStoreConfig.builder(rv.SEMANTIC_SEGMENTATION_RASTER) \
                                          .with_vector_output([vector_output]) \
                                          .with_rgb(True) \
--- a/integration_tests/semantic_segmentation_tests/experiment.py
+++ b/integration_tests/semantic_segmentation_tests/experiment.py
@@ -55,9 +55,16 @@ class SemanticSegmentationIntegrationTest(rv.ExperimentSet):
                                            .with_raster_source(label_path) \
                                            .build()
 
-        vector_output = {'denoise': 1, 'mode': 'buildings', 'class_id': 1}
+        vector_output = [{
+            'mode': 'buildings',
+            'class_id': 1
+        }, {
+            'denoise': 50,
+            'mode': 'polygons',
+            'class_id': 1
+        }]
         label_store = rv.LabelStoreConfig.builder(rv.SEMANTIC_SEGMENTATION_RASTER) \
-                                         .with_vector_output([vector_output]) \
+                                         .with_vector_output(vector_output) \
                                          .with_rgb(True) \
                                          .build()
 
--- a/integration_tests/semantic_segmentation_tests/experiment.py
+++ b/integration_tests/semantic_segmentation_tests/experiment.py
@@ -55,7 +55,9 @@ class SemanticSegmentationIntegrationTest(rv.ExperimentSet):
                                            .with_raster_source(label_path) \
                                            .build()
 
+        vector_output = {'uri': '*', 'mode': 'buildings'}
         label_store = rv.LabelStoreConfig.builder(rv.SEMANTIC_SEGMENTATION_RASTER) \
+                                         .with_vector_output([vector_output]) \
                                          .with_rgb(True) \
                                          .build()
 
--- a/integration_tests/semantic_segmentation_tests/experiment.py
+++ b/integration_tests/semantic_segmentation_tests/experiment.py
@@ -37,7 +37,7 @@ class SemanticSegmentationIntegrationTest(rv.ExperimentSet):
                                                debug_chip_probability=1.0) \
                             .build()
 
-        # .with_config belowe needed to copy final layer from pretrained model.
+        # .with_config below needed to copy final layer from pretrained model.
         backend = rv.BackendConfig.builder(rv.TF_DEEPLAB) \
                                   .with_task(task) \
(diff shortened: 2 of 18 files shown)


lewfish:
I haven't looked at this in any detail yet, but one request is to add `polygon` mode which extracts polygons without doing the extra right-angle stuff for buildings.

integration_tests/semantic_segmentation_tests/experiment.py
```diff
@@ -55,7 +55,9 @@ def get_path(part):
                                            .with_raster_source(label_path) \
                                            .build()
 
+        vector_output = {'uri': '*', 'mode': 'buildings'}
```
lewfish: Rather than use '*' to indicate that it should be autogenerated, I think `uri` should have a default value of None which is treated as "autogenerate" to be more consistent with the rest of RV.

rastervision/data/label_store/semantic_segmentation_raster_store.py
```diff
@@ -96,6 +110,20 @@ def save(self, labels):
...
 
+        for vo in self.vector_output:
+            uri = vo['uri']
+            mode = vo['mode']
+            local_geojson_path = get_local_path(uri, self.tmp_dir)
+            if mode == 'buildings':
+                import mask_to_polygons.vectorification as m2p
+
+                with rasterio.open(local_path, 'r') as dataset:
+                    mask = dataset.read(1)
```
lewfish: This only will work with single band GeoTIFF output (as opposed to RGB output). Maybe we could avoid this issue by just doing the `m2p` earlier in this method when we have the raw predictions in RAM in single band format. Also, does this work with multiple classes? What assumptions does it make about the class_ids?

lewfish: If it's going to be a lot of work to make it more flexible you can make a separate issue for it, but just make sure to document the limitations and add input validation in this PR (ie. check that the vector output config is consistent with the `with_rgb` setting and the `class_map`

rastervision/data/scene_config.py
```diff
@@ -36,6 +36,12 @@ def create_scene(self, task_config: TaskConfig, tmp_dir: str) -> Scene:
         raster_source = self.raster_source.create_source(tmp_dir)
 
         extent = raster_source.get_extent()
+        if self.label_store and hasattr(
+                self.label_store,
+                'vector_output') and self.label_store.vector_output:
+            task_config.affine_transform = raster_source.get_affine_transform()
```
lewfish: It seems like the `task_config` is an odd place to stick the `affine_transform` since it's something that's associated with the raster source in the scene and not the task (which is constant across a set of tasks).  Instead, maybe you can just pass the `affine_transform` to the label_store directly.  It also might make sense to just make `affine_transform` a part of the crs_transformer, but I'm not sure about that. It would eliminate the issue of having to pass an extra thing along, and it fits with the role of the crs_transformer, but I haven't considered all the consequences of that change.

rastervision/data/label_store/semantic_segmentation_raster_store_config.py
```diff
@@ -64,27 +92,67 @@ def update_for_command(self,
...
         """Set URI for a GeoTIFF used to read/write predictions."""
         b = deepcopy(self)
         b.config['uri'] = uri
         return b
 
+    def with_vector_output(self, msg):
+        """Configure vector output for predictions.
+
+            Args:
+                msg: Either a list of dictionaries or a protobuf
```
lewfish: I'm pretty sure that none of the other `with` methods take protobuf messages as input (with the exception of with_class_map). I guess it doesn't hurt, but it doesn't seem very useful. I also think the name `config` is better than `msg` since it might be a dict . Also, I would add something to the validate method to check that the value of mode is valid.

Dockerfile
```diff
@@ -6,13 +6,14 @@ FROM tensorflow/tensorflow:1.10.1-gpu-py3
...
 RUN add-apt-repository ppa:ubuntugis/ppa && \
     apt-get update && \
     apt-get install -y wget=1.* git=1:2.* python-protobuf=2.* python3-tk=3.* \
-            gdal-bin=2.1.* \
-            jq \
-            build-essential libsqlite3-dev zlib1g-dev && \
+                       gdal-bin=2.1.* \
+                       jq \
+                       build-essential libsqlite3-dev zlib1g-dev \
+                       libopencv-dev python-opencv && \
```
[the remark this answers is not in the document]
jamesmcclain: Primarily because they were not pinned before: https://github.com/azavea/raster-vision/blob/ba83d57a1b1b65fe8b5693ed935f22804723fba3/Dockerfile#L27 https://github.com/azavea/raster-vision/blob/ba83d57a1b1b65fe8b5693ed935f22804723fba3/Dockerfile#L52-L54 .  Not a problem.

rastervision/data/scene_config.py
```diff
@@ -36,6 +36,7 @@ def create_scene(self, task_config: TaskConfig, tmp_dir: str) -> Scene:
         raster_source = self.raster_source.create_source(tmp_dir)
 
         extent = raster_source.get_extent()
+        affine_transform = raster_source.get_affine_transform()  # XXX
```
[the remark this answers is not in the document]
jamesmcclain: Must have been a stray comment.  I cannot find it anymore.

rastervision/data/label_store/semantic_segmentation_raster_store.py
```diff
@@ -12,11 +12,14 @@ class SemanticSegmentationRasterStore(LabelStore):
     """A prediction label store for segmentation raster files.
     """
 
-    def __init__(self, uri, extent, crs_transformer, tmp_dir, class_map=None):
+    def __init__(self, uri, geojson_uri, extent, affine_tranform, crs_transformer, tmp_dir, class_map=None):
         """Constructor.
 
         Args:
             uri: (str) URI of GeoTIFF file used for storing predictions as RGB values
+            geojson_uri: (str or None) URI of GeoJSON polygons derived from the mask
```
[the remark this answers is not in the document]
jamesmcclain: Resolved.

rastervision/data/label_store/semantic_segmentation_raster_store.py
```diff
@@ -12,11 +12,14 @@ class SemanticSegmentationRasterStore(LabelStore):
...
+            extent: (Box) The extent of the scene
+            affine_tranform: (affine.Affine) The mapping from mask coordintes to world coordinates
```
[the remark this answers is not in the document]
jamesmcclain: Resolved.

integration_tests/semantic_segmentation_tests/experiment.py
jamesmcclain: Resolved.

rastervision/data/label_store/semantic_segmentation_raster_store.py
jamesmcclain: Done.

rastervision/data/label_store/semantic_segmentation_raster_store_config.py
```diff
@@ -64,27 +98,71 @@ def update_for_command(self,
...
+
+        if isinstance(vector_output, list):
+            for vo in vector_output:
+                ar.append(vo.copy())
+        else:
+            for vo in vector_output:
+                ar.append({
+                    'uri': vo.uri,
+                    'mode': vo.mode,
+                    'class_id': vo.class_id
```
lewfish: This produces a separate file for each class/mode. That has the advantage of allowing you to use several different modes for each class. I could see how that could be useful. On the other hand, it produces a file(s) for each class, rather than a single file where each feature has a `class_id` field, which is how things work in other parts of RV. I think it's ok the way it is, but I just wanted to note this tradeoff.

rastervision/data/label_store/semantic_segmentation_raster_store.py
```diff
@@ -126,7 +128,11 @@ def save(self, labels):
                 class_mask = np.array(mask == class_id, dtype=np.uint8)
                 local_geojson_path = get_local_path(uri, self.tmp_dir)
 
-                transform = Affine.identity()
+                if isinstance(self.crs_transformer, RasterioCRSTransformer):
```
lewfish: If someone adds a new `CRSTransformer`, they might not remember to update this code which would end up returning an identity transform when that might not be appropriate.

Instead, you should add a method `get_affine_transform` to `CRSTransformer` which raises `NotImplementedError` as an exception. Then subclasses can implement their own way of generating and returning the affine transform.

lewfish: The reason I decided it shouldn't be an abstract method is that then subclasses would be forced to implement it, which seems like overkill since it's only used in this one place, and writers of new raster sources and crstransformer might not be interested in feature extraction.

jamesmcclain: Okay.

lewfish:
I tested this using a modification of the Vegas example in 
https://github.com/azavea/raster-vision-examples/commit/65417f60612bfad0f9c8782d185907bebe1b92c4

The command run was 
```
export ROOT_URI=/opt/data/lf-dev/rv-output/vegas-spacenet
rastervision run local -e spacenet.vegas \
    -a test True \
    -a use_remote_data False \
    -a root_uri ${ROOT_URI} \
    -a target buildings \
    -a task_type semantic_segmentation
```
If you don't have the Vegas data on your local filesystem you can switch use_remote_data to True

lewfish:
Note that I updated the branch/commit that the Vegas test is in above.

jamesmcclain:
I have one non-empty geojson for each of the four scenes:

screenshot from 2018-12-18 12-46-00

lewfish:
> I have one non-empty geojson for each of the four scenes:

Does that mean that it's working the way you expect?

jamesmcclain:
> Does that mean that it's working the way you expect?

Most probably.  You seemed to be describing either empty geojson files or none at all?

lewfish:
> Most probably. You seemed to be describing either empty geojson files or none at all?

I had empty files before. So this seems like an improvement. You can merge this.

jamesmcclain:
I didn't make any improvements ...

jamesmcclain:
With https://github.com/jamesmcclain/raster-vision/pull/1

## Ground truth ##
screenshot_2018-12-21_10-24-40

## Predictions ##
screenshot_2018-12-21_10-24-30

## Eval ##
The file `/tmp/XXX/eval/buildings_semantic_segmentation/eval.json` should look something like this:

```json
[
    {
        "class_id": 1,
        "count_error": 399941.64566854003,
        "precision": 0.03893706315187772,
        "class_name": "Building",
        "f1": 0.06278322228817987,
        "gt_count": 303916,
        "recall": 0.16918819673857252
    },
    {
        "class_id": 2,
        "count_error": 401015.0285372315,
        "precision": 0.12290437362221147,
        "class_name": "Background",
        "f1": 0.04184535492697485,
        "gt_count": 1386084,
        "recall": 0.02546238178927107
    },
    {
        "class_id": -2,
        "count_error": 13.26605504587156,
        "precision": 0.9267246514952938,
        "class_name": "vector-polygons",
        "f1": 0.7137699626231737,
        "gt_count": 109,
        "recall": 0.5871559633027523
    },
    {
        "class_id": null,
        "count_error": 400796.1506896892,
        "precision": 0.10785718274954362,
        "class_name": "average",
        "f1": 0.04565374401855821,
        "gt_count": 1690109,
        "recall": 0.051343434062536795
    }
]
```

Issue #583, opened 2018-11-29
Improve local runner


lewfish:
The local runner was changed to be more like the remote runner in https://github.com/azavea/raster-vision/pull/572 There are a few problems that I didn't recognize during code review which should be fixed.
* It seems odd that the local runner now inherits from AwsBatchRunner
* Having `Old...` in the name of something is confusing especially as time goes on.
* When one step errors now, all the other commands still carry through (unlike the aws batch runner, which cancels commands if a parent one fails). This may or may not be hard to fix given the inheritance structure.

PR #588, opened 2018-11-30
[WiP] Improvements to Local Runner

jamesmcclain:
## Overview

Pursuant to https://github.com/azavea/raster-vision/issues/583

### Checklist

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

Closes https://github.com/azavea/raster-vision/issues/583


tests/mock/raster_transformer.py
```diff
@@ -92,7 +92,10 @@ def __init__(self, prev=None):
     def from_proto(self, msg):
         result = self.mock.from_proto(msg)
         if result is None:
-            return super().from_proto(msg)
+            retval = super().from_proto(msg)
+            if not retval and msg.transformer_type == 'MOCK_TRANSFORMER':
+                retval = MockRasterTransformerConfigBuilder()
+            return retval
         else:
```
jamesmcclain: @simonkassel This may be related to the issue that you were seeing.

rastervision/runner/out_of_process_experiment_runner.py
```diff
@@ -0,0 +1,80 @@
...
+                    upstream_command = (u.command_type, upstream_id)
+                    raise Exception(
+                        '{} command has parent command of {}, '
+                        'but does not exist in previous batch submissions - '
+                        'topological sort on command_dag error.'.format(
+                            cur_command, upstream_command))
+                parent_job_ids.append(ids_to_job[upstream_id])
+
+            batch_run_command = make_command(command_uri, self.tmp_dir)
+            job_id = self.submit(
```
lewfish: I don't think the `OutOfProcess..` should call a `submit` method that passes a `job_queue` and `job_def` since that is too tied to Batch concepts. A better way might be to call a `run_command` method that omits the `job_queue`, `job_def` and `attempts`. (I'll admit that `run_command` isn't perfect either since it sort of implies that something will happen synchronously, when in fact it could be asynchronous in the case of Batch, so I'm ok with leaving it as `submit` if you like that better). These don't need to be passed anyway since they are attached to `self`. I don't see anyway around passing the `parent_job_ids` though.

rastervision/runner/out_of_process_experiment_runner.py
```diff
@@ -0,0 +1,80 @@
...
+    if tmp_dir is None:
+        command = 'python -m rastervision {} run_command {}'.format(
+            v_flag, command_config_uri)
+    else:
+        command = 'python -m rastervision {} run_command {} --tempdir {}'.format(
+            v_flag, command_config_uri, tmp_dir)
+    return command
+
+
+class OutOfProcessExperimentRunner(ExperimentRunner):
```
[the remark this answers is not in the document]
jamesmcclain: Done

rastervision/runner/out_of_process_experiment_runner.py
jamesmcclain: I am not sure that I understand what is written here, but I have turned the respective submit functions into class methods which removes the need for some parameters.

Issue #571, opened 2018-11-19
Support Multiple GPUs in Deeplab


jamesmcclain:
In the case of the deeplab backend, it is fairly simple.  We need to add the switch `--num-clones` followed by the number of GPU devices to the training command line.  This should perhaps be controlled by configuration and/or a rastervision command line argument.

See also https://github.com/azavea/raster-vision/issues/563
Motivated by https://gitter.im/azavea/raster-vision?at=5bf2e1a0f9993b42673491f8

PR #590, opened 2018-11-30
Add Multi-GPU Support to DeepLab Backend

jamesmcclain:
## Overview

Add multi-GPU support to the DeepLab backend.

### Checklist

Closes https://github.com/azavea/raster-vision/issues/571


rastervision/backend/model_defaults.json
```diff
@@ -97,6 +97,7 @@
                 "trainBatchSize": 24,
                 "trainCropSize": [257, 257],
                 "trainingNumberOfSteps": 35410,
+		"numClones": 1,
```
lewfish: `numClones` is not indented the same way as the other fields (here and the instance below).

lewfish:
To test this we'll need to use a multi-GPU machine. I made a separate issue for this and some other things. https://github.com/azavea/raster-vision/issues/600

rastervision/backend/model_defaults.json
jamesmcclain: Needed to be untabified.

PR #617, opened 2018-12-12
Handle multiple AOI URIs

### LINKED ISSUE #615: Handle multiple AOI URIs
It is convenient to have multiple AOIs associated with a scene stored in different files. This way the AOIs used for training / testing could be set by the experiment configuration script. However, it's not easy to do this at the moment because you can only set a single AOI file using `with_aoi_uri`. If we had a `with_aoi_uris` method, then we could handle this.

lewfish:
## Overview

This PR adds a `with_aoi_uris` method to `SceneConfigBuilder` to handle multiple files containing AOIs. This makes it easier to select arbitrary subsets of AOIs to associate with a scene when experimenting. The original `with_aoi_uri` method, and `aoi_uri` `Scene` protobuf fields are still intact and functional for backward compatibility, 

Closes #615

### DIFF
--- a/rastervision/protos/scene.proto
+++ b/rastervision/protos/scene.proto
@@ -22,5 +22,9 @@ message SceneConfig {
     optional LabelStoreConfig prediction_label_store = 4;
 
     // Optional AOI that describes where the scene is fully labeled.
+
+    // XXX here for backward compatibility
     optional string aoi_uri = 5;
+    
+    repeated string aoi_uris = 6;
 }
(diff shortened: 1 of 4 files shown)


rastervision/protos/scene.proto
```diff
@@ -22,5 +22,9 @@ message SceneConfig {
     optional LabelStoreConfig prediction_label_store = 4;
 
     // Optional AOI that describes where the scene is fully labeled.
+
+    // XXX here for backward compatibility
     optional string aoi_uri = 5;
```
jamesmcclain: XXX [may not be called for](https://softwareengineering.stackexchange.com/questions/65467/what-does-xxx-mean-in-a-comment) here.

lewfish: True, I took it out.

PR #619, opened 2018-12-12
Use Popen to Run Shell Commands.

### LINKED ISSUE #609: Ctrl-c does not kill child commands
It's a little hard to replicate this, but sometimes when running commands with the local runner, hitting ctrl-c kills the process, but not the child commands which were created by shelling out. I'm pretty sure this is due to the changes we just made to the command runner. We need to do the same thing we do in the backends when shelling out to avoid this.

jamesmcclain:
## Overview

Use Popen to run shell commands.

### Checklist

Closes #609


jamesmcclain:
Okay to merge.

PR #657, opened 2019-01-16
Parallelize Chipping Stage

jamesmcclain:
## Overview

Brief description of what this PR does, and why it is needed.

### Checklist

- [x] Updated `docs/changelog.rst`
- [x] Added `needs-backport` label if PR is bug fix that applies to previous minor release
- [x] Ran scripts/format_code and committed any changes
- [x] Documentation updated if needed
- [x] PR has a name that won't get you publicly shamed for vagueness
- [x] Run processes in parallel where possible locally
- [x] Test on AWS

## Testing Instructions

Add the `--splits` flag (along with some integer) to the command line of some chosen experiment.

Connects #654 
Depends on #661
Depends on #664

PR #664, opened 2019-01-18
Use `make` to Manage Local Execution

jamesmcclain:
## Overview

GNU `make` is now used for managing local execution.  This is a prerequisite for having parallel, simultaneous processes with dependency relationships between them.  There should be no discernible difference in behavior with only this change but without stage splitting -- with stage splitting (https://github.com/azavea/raster-vision/pull/657) independent stages and parts of stages run in parallel.

GNU make is the default on virtually all Linux distributions and on OSX.  Windows users are invited to install [MinGW's version](http://www.mingw.org/wiki/msys) or to use the `inprocess` runner for local jobs.

### Checklist

### DIFF
--- a/rastervision/runner/local_experiment_runner.py
+++ b/rastervision/runner/local_experiment_runner.py
@@ -1,9 +1,13 @@
 from subprocess import Popen
 import sys
 import logging
+import os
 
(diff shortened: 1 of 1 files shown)


jamesmcclain:
Thanks

PR #661, opened 2019-01-17
Default Values in the In-Process Runner

jamesmcclain:
## Overview

Some default values are supplied by Protobuf.  This change allows experiments that rely on that behavior to succeed with this runner (namely the semantic segmentation integration test).

### Checklist

- [x] Updated `docs/changelog.rst`
- [x] Added `needs-backport` label if PR is bug fix that applies to previous minor release
- [x] Ran scripts/format_code and committed any changes
- [x] Documentation updated if needed
- [x] PR has a name that won't get you publicly shamed for vagueness

## Testing Instructions

Run the semantic segmentation integration test in `inprocess` mode with and without these changes.

PR #667, opened 2019-01-24
Parallelize Prediction Stage

### LINKED ISSUE #654: Allow splitting of commands for parallel execution
Develop a method for users to specify the number of parallelizable components a command configuration can split itself into. This will allow us to do the following: if we are going through the CHIP, TRAIN, PREDICT, EVAL, and BUNDLE steps, and we want to use a split size of 5, the following will happen:
- 5 CHIP command configs will be generated, each containing 20 images. These will be executed in parallel based on the runner's ability to execute independent commands in parallel - e.g. AWS Batch would spin up 5 instances to perform the 5 chip commands.
- 1 TRAIN command config will be generated - TRAIN is not able to split itself into parallelizable components, so the split option has no effect on this command.
- 5 PREDICT commands configs are generated
- 5 EVAL commands configs are generated
- 1 BUNDLE command is generated.

A command line invocation of the split run could look like:

```
> rastervision run aws_batch -e some.experiment --split 5
```

or

```
> rastervision run aws_batch -e some.experiment --s 5
```

A sketch of accomplishing this task is as follows:
- Add an option to the run command in `cli.main.py` for the split option, which gets pass to the runner
- The runner will use the split option to optionally call a `.split(split_number)` on each command in `unique_commands` [before the DAG is created](https://github.com/azavea/raster-vision/blob/697c791beef75b4adb1f341aefd47ae5a66fc1b0/rastervision/runner/experiment_runner.py#L152), constructing a new list from the results:

```
unique_commands = [command_config  for cc in unique_commands for command_config in cc.split(split_number)]
```

- A `.split(desired_number_of_splits)` method is added to `CommandConfig`, which returns itself in a list if it cannot be split, or returns a set of split commands that replace the current command (e.g. the 5 CHIP commands in the example above)
- The CommandConfigs that can split themselves call a `.split` method on the constituent Configs, so `split` methods would have to be added to the base `Config` class and by default return `self`. For example, `Dataset.split(n)` would split itself by each of the train, validation and test sets. It would be up to the CHIP, EVAL and PREDICT commands to throw out any datasets that only contained train scenes, for example.

`split` on `Config` will need to do more than just split data - it will need to change output paths so that split parts do not write over each other. It also has to account for input paths in downstream commands that are determined by output paths from commands. Figuring out this bit will be a challenge of this task.

Continuation of #428 and #551

jamesmcclain:
## Overview

These changes allow multiple predictions to run in parallel.

### Checklist

### Notes

Add the `--splits` flag (along with some integer) to the command line of some chosen experiment.

Closes #654 
Depends on #661
Depends on #664
Depends on #657

### DIFF
--- a/rastervision/command/chip_command_config.py
+++ b/rastervision/command/chip_command_config.py
@@ -103,11 +103,11 @@ class ChipCommandConfigBuilder(CommandConfigBuilder):
                         'val_scenes must be a list of class SceneConfig, '
                         'got a list of {}'.format(type(s)))
 
-    def build(self):
+    def build(self, index: int = 0, count: int = 1):
         self.validate()
-        return ChipCommandConfig(self.root_uri, self.task, self.backend,
-                                 self.augmentors, self.train_scenes,
-                                 self.val_scenes)
+        return ChipCommandConfig(
+            self.root_uri, self.task, self.backend, self.augmentors,
+            self.train_scenes[index::count], self.val_scenes[index::count])
 
     def from_proto(self, msg):
         b = super().from_proto(msg)
--- a/rastervision/command/chip_command_config.py
+++ b/rastervision/command/chip_command_config.py
@@ -103,11 +103,26 @@ class ChipCommandConfigBuilder(CommandConfigBuilder):
                         'val_scenes must be a list of class SceneConfig, '
                         'got a list of {}'.format(type(s)))
 
-    def build(self, index: int = 0, count: int = 1):
+    def build(self, io_def=None):
         self.validate()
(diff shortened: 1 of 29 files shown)


rastervision/command/chip_command_config.py
```diff
@@ -103,11 +103,26 @@ def validate(self):
                         'val_scenes must be a list of class SceneConfig, '
                         'got a list of {}'.format(type(s)))
 
-    def build(self):
+    def build(self, io_def=None):
```
jamesmcclain: .

Issue #649, opened 2019-01-03
Add support for running some jobs on CPUs


lewfish:
When using Batch, we would like to be able to run some jobs on the CPU and some on GPU. To do this, we will need to modify the RV profile configuration to specify CPU and GPU resources. Currently a config looks like:
```
[AWS_BATCH]
job_queue=raster-vision-lewfish-p3
job_definition=raster-vision-lewfish-p3
```
but it should become something like
```
[AWS_BATCH]
cpu_job_queue=raster-vision-lewfish-m4
cpu_job_definition=raster-vision-lewfish-m4
gpu_job_queue=raster-vision-lewfish-p3
gpu_job_definition=raster-vision-lewfish-p3
```

Then the `aws_batch` runner needs to be modified so that it picks the correct job queue and job def for the command. Each command needs to have a mechanism for deciding which queue should be used (analyze, chip, and eval seem like good candidates for CPU). When jobs are submitted, the required number of CPU cores needs to be specified. It should stay at the current value for GPUs, but CPU jobs should only require one core.

A prerequisite for this issue is https://github.com/azavea/raster-vision/issues/634

Issue #634, opened 2018-12-20
Check if cross queue dependencies are possible


lewfish:
Around a year ago it wasn't possible to have a job in one batch queue have a dependency on a job in another queue. This is unfortunate because we would like to use CPU instances for some jobs and GPU instances for others. We should do a little test and/or reading to see if this is possible now.

PR #668, opened 2019-01-25
Make Use of CPU and GPU Queues

jamesmcclain:
## Overview

### Checklist

Closes https://github.com/azavea/raster-vision/issues/634
Closes https://github.com/azavea/raster-vision/issues/649

### DIFF
--- a/rastervision/runner/aws_batch_experiment_runner.py
+++ b/rastervision/runner/aws_batch_experiment_runner.py
@@ -24,6 +24,14 @@ class AwsBatchExperimentRunner(OutOfProcessExperimentRunner):
                 job_queue = 'raster-vision-cpu'
         self.job_queue = job_queue
 
+        cpu_job_queue = batch_config('cpu_job_queue', default='')
+        if not cpu_job_queue:
+            if self.gpu:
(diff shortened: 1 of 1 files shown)


lewfish:
I thought this worked but when I looked at the Batch console I noticed that the first job is stuck in Runnable. This could be because there's something messed up with the new Batch resources I just created using the new CloudFormation setup. But it also looks like what happened in the past when we had jobs with cross-queue dependencies. When you tested whether this was possible, did you notice if the jobs were actually completed?

jamesmcclain:
> did you notice if the jobs were actually completed?

All completed.

lewfish:
After making some changes (for one, lowering the requested RAM) I've got the jobs to move past Runnable in the CPU queue but they still crash. I think there's something wrong with the Cloudformation setup. I have one more idea to try before I contact Ops.

jamesmcclain:
> After making some changes (for one, lowering the requested RAM) I've got the jobs to move past Runnable in the CPU queue but they still crash. I think there's something wrong with the Cloudformation setup. I have one more idea to try before I contact Ops.

Okay

PR #671, opened 2019-01-30
Parallelization of commands (CHIP and PREDICT)

### LINKED ISSUE #268: Travis: Avoid double building on PR's originating from this repository
Currently, Travis will double build PR's that are created and pushed to. One is because the `Build pushed branches` option is on, and the other is because the `Build pushed pull requests` are on. We want the latter on because we should be building PRs that may or may not originate from branches in this repo; we want to former because otherwise the `develop` branch and tags won't build.

I scratched my head on this for a while but am putting it away as it's more a nuisance than a bug. Here are some of the resources I was looking at to figure out the right configuration:

https://docs.travis-ci.com/user/pull-requests/#%E2%80%98Double-builds%E2%80%99-on-pull-requests
https://docs.travis-ci.com/user/web-ui/#Build-pushed-branches
https://docs.travis-ci.com/user/customizing-the-build/#Building-Specific-Branches

Note in the `Building Specific Branches` section, building of tags will be disabled unless we use a regular expression to release tags that then have to be named consistently.

### LINKED ISSUE #654: Allow splitting of commands for parallel execution

lossyrob:
This PR adds the ability for users to specify a `split` option for running experiments, which will split certain commands into smaller commands that can be run in parallel.

This introduces several refactors:
- Utilizing directories instead of specific file paths in the CommandIODefinitions
- Breaking up the  `update_for_command` from the addition of files to io_defs, which is now done in `report_io`.

This was tested against Rio chip classification and Vegas semantic segmentation in the examples.

TODO
- [x] Increase unit test coverage
- [x] Add documentation
- [ ] ~~Modify examples to use splits~~ Examples need to be updated further as per https://github.com/azavea/raster-vision-examples/issues/42 - Adding splitting can be done as part of the upgrade after this gets merged.
- [x] Add release notes
- [x] Document `report_io` and `update_for_command`, in general and specifically for plugins.

Closes #654 
Closes #268


rastervision/runner/command_definition.py
```diff
@@ -36,12 +40,14 @@ def from_experiments(cls, experiments: List[rv.ExperimentConfig]):
...
                 log.debug('Creating experiment configuration...'.format(
                     command_type))
-                command_config = e.make_command_config(command_type)
-                command_def = cls(e.id, command_config, io_def)
-                command_definitions.append(command_def)
+                base_command_config = e.make_command_config(command_type)
+                for command_config in base_command_config.split(splits):
+                    io_def = command_config.report_io()
+                    command_def = cls(e.id, command_config, io_def)
+                    command_definitions.append(command_def)
```
lossyrob: This is the  crux of the changes - by separating the io_def updates and the update_for_command, we can generate commands from the fully resolved and unsplit experiment, and then split the commands, then ask the commands to provide the io_def updates.

lewfish:
I skimmed this and the approach seems sound.

rastervision/runner/aws_batch_experiment_runner.py
```diff
@@ -75,13 +86,40 @@ def batch_submit(self,
...
 
         client = boto3.client('batch')
 
         uuid_part = str(uuid.uuid4()).split('-')[0]
         exp = ''.join(e for e in experiment_id if e.isalnum())
-        job_name = '{}_{}_{}'.format(command_type, exp, uuid_part)
+        job_name = '{}-{}_{}_{}'.format(command_type, command_split_id, exp,
+                                        uuid_part)
+
+        group_level = 1
```
lossyrob: @lewfish commented in slack:

You can use Batch "array jobs" for this. https://docs.aws.amazon.com/batch/latest/userguide/array_jobs.html There's some code for this here https://github.com/azavea/raster-vision/blob/5ce55eef38abfc774779a9c93ccadca7e905d73a/rastervision/runner/aws_batch_experiment_runner.py#L107

lossyrob: We'd have to figure out how to utilize AWS_BATCH_JOB_ARRAY_INDEX inside `run_command`, as this is the only mechanism for splitting up the command parameters AFAIK. https://github.com/azavea/raster-vision/blob/f62810a99f0c5f00e1467163130e01ea05c757ea/rastervision/cli/main.py#L247

We could feed an option into run_command that specifies the environment variable that defines the split ID, and then modify the command config appropriately. Thoughts?

lossyrob: This will also affect the mechanism for submission - the submission happens a command at a time, and does not have the visibility of e.g. all CHIP commands at once. This may take a substantial refactor to work out - and may take some reconsideration of the OutOfProcessRunner heirarchy.

Perhaps we should do this in a separate PR?

lewfish: Yeah, that would be a big change. I think the way you have it is ok.

lewfish:
When testing this on a remote object detection experiment, it fails due to `self.temp_dir` never getting set https://github.com/azavea/raster-vision/blob/3d9835285ede8fb45722c0d7e81bef304d1a7b9d/rastervision/backend/tf_object_detection.py#L396

rastervision/command/chip_command_config.py
```diff
@@ -68,7 +69,8 @@ def split(self, num_parts):
         commands = []
         t_scenes = list(map(lambda x: (0, x), self.train_scenes))
         v_scenes = list(map(lambda x: (1, x), self.val_scenes))
-        group_size = max(int((len(t_scenes) + len(v_scenes)) / num_parts), 1)
+        group_size = max(
```
lewfish: It seems like the logic for getting group size should be moved into a utility function since it is duplicated in `predict_command_config`. You may also want to consider combining that and the `grouped` function into a single `split_into_groups(list, num_groups)` since they usually co-occur. In addition, I think this function should be unit tested so we can make sure it is correct in edge cases, like more groups than elements in list, number of elements isn't divisible by number groups, etc.

.travis.yml
```diff
@@ -11,6 +11,8 @@ env:
       - CLEAN_TRAVIS_TAG=${TRAVIS_TAG/[[:space:]]/}
       - COMMIT=${CLEAN_TRAVIS_TAG:-${TRAVIS_COMMIT:0:7}}
 
+if: (type = pull_request) OR (tag IS present) OR (branch = develop) OR (branch =~ /^feature.*/)
```
lewfish: Does this fix https://github.com/azavea/raster-vision/issues/268?

lossyrob: Yep! I'll mark it in the PR description.

rastervision/command/command_config.py
```diff
@@ -24,6 +25,13 @@ def report_io(self):
         """
         pass
 
+    def split(self, num_parts):
+        """Split this command config into num_parts parts if possible.
+        This will return a list containing 1 - num_parts command configurations.
```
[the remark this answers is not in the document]
lossyrob: Ah, should say "1 to num_parts", I'll reword it.

PR #676, opened 2019-01-31
Optimize vector tile processing

### LINKED ISSUE #557: Improve vector tile performance
There are a few ways to improve vector tile performance if needed:
- [x] Download vector tile files to cache
- [x]  Filter features before doing merge and  crop step
- [x] In-memory cache (ideally LRU)

lewfish:
## Overview

This PR changes the way vector tiles are processed to be more efficient. The two main improvements are: caching the GeoJSON that is extracted from a tile (since many scenes may use the same tile, especially at low zoom levels) and doing the feature merge after irrelevant features are filtered out. 

## Testing
I tested this on the Vegas roads example using the following and found that the debug chips looked correct. I then timed the run by prefixing the invocation with `time` and also set `debug=False` in `spacenet.vegas` (to avoid timing the part where it's producing debug chips, which is not relevant to the changes in this PR). I repeated this 3 times on this branch, and 3 times on `develop` and found that this PR decreased the run time by ~25%. That's not the dramatic speedup I was hoping for, but maybe this example doesn't do the best job of utilizing the improvements. 

```
rastervision run local -e spacenet.vegas \
    -a test True \
    -a use_remote_data False \
    -a root_uri /opt/data/spacenet/vegas/ \
    -a target roads \
    -a task_type semantic_segmentation \
    -a vector_tile_options "https://s3.amazonaws.com/osmesa-vectortiles/full/2018-08-02/{z}/{x}/{y}.mvt,14,__id" \
    -r chip
```

Closes #557


rastervision/data/vector_source/vector_tile_vector_source.py
```diff
@@ -45,23 +43,43 @@ def process_features(features, map_extent, id_field):
...
-    return proc_features
 
+    return {'type': 'FeatureCollection', 'features': proc_features}
 
-def vector_tile_to_geojson(uri, zoom, id_field, crs_transformer, extent):
-    """Get GeoJSON covering an extent from a vector tile endpoint.
 
-    Merges features that are split across tiles and crops against the extentself.
-    """
+tile_feature_cache = {}
```
jamesmcclain: Is there a replacement policy?  Are there concerns about scalability (this is a question not a loaded question)?

lewfish: I'm not sure, but it's a better idea to set some limit on the size of the cache. Python 3 has a nice `lru_cache` callback that I'm using now. I'm not sure what a good limit size is for the cache, but 32 seems safe.

jamesmcclain: > I'm not sure, but it's a better idea to set some limit on the size of the cache. Python 3 has a nice `lru_cache` callback that I'm using now. I'm not sure what a good limit size is for the cache, but 32 seems safe.

That sounds like a fine plan.

PR #690, opened 2019-02-19
Optimize StatsAnalyzer 🏁

### LINKED ISSUE #648: Optimize StatsAnalyzer/RasterStats logic
The StatsAnalyzer allows users to get statistics of their raster data that can then be used by the StatsTransformer to rescale the data from wider datatypes to uint8.

The StatsAnalyzer however is *very* slow. This issue is to optimize the logic.

The StatsAnalyzer uses the [RasterStats](https://github.com/azavea/raster-vision/blob/1b2543ac3ac261352bba07403dd57c106cac731b/rastervision/core/raster_stats.py)  to compute the statistics. There are a couple of bottlenecks here:
- The logic uses a chip stream to compute the means, and then the standard deviation - the std calculation recalculates the mean, so the chip collection is iterated over an additional time unnecessarily.
- The stats are computed over the entire dataset, when an approximation could take a sample of the data and perform adequately.

This issue should solve the performance issues with StatsAnalyzer to increase speed by addressing the above issues and potentially exploring other ways to optimize the logic.

lewfish:
## Overview

This PR speeds up the `StatsAnalyzer` (which computes means and standard deviations of each channel in the dataset) by only getting each chip once rather than `2*nb_channels` times, switching away from `npstreams`, and adding an option for approximating the stats using a random subset of chips. It also adds unit tests for the StatAnalyzer. Example:
```python
        analyzer = rv.AnalyzerConfig.builder(rv.STATS_ANALYZER) \
                                    .with_sample_prob(0.01) \
                                    .build()

        experiment = rv.ExperimentConfig.builder() \
                                        .with_id('potsdam-seg') \
                                        .with_analyzer(analyzer) \
```

### Notes
I thought about making some `sample_prob` the default, but decided against it because I don't know a good value yet and the value could depend on the average size of a scene. On the other hand, it seems like the fast option should be the default. But even with sliding window mode (ie. no sample_prob) it still runs 6x faster.

Another issue is that using the `sample_prob` option will result in greater approximation error if the scenes contain a large amount of NODATA.

### Testing
The unit tests exercise the new functionality. I also benchmarked it on Potsdam with two scenes and it resulted in a huge speedup (> 30x) with only a small difference in the stats that were computed.

Using `develop`:
time 2:07
```
{"stds": [35.870648282240516, 35.32982639561334, 36.702463479669596, 37.38938087670711], "means": [85.95280349150948, 92.19038055254074, 85.52430486264436, 96.83653263423753]}
```

Using this branch with no `sample_prob`: 
time 0:21
```
{"stds": [35.870648282240516, 35.32982639561334, 36.702463479669596, 37.38938087670711], "means": [85.95280349150948, 92.19038055254074, 85.52430486264436, 96.83653263423753]}
```

Using this branch with `sample_prob=0.01`: 
time 0:03
```
{"stds": [36.17572923393826, 35.857899881444986, 37.11680588294381, 36.18744804702934],
"means": [89.91624927520753, 96.44924354553223, 87.8176760673523, 111.6387071609497]}
```

Closes #648


ljvmiranda921:
Thank you for this, this has been really helpful!

PR #754, opened 2019-04-09
Switch from model to exported-model

lewfish:
## Overview

This PR switches from naming the exported model from `model` to `exported-model`. This is needed because `file_exists`, when used on S3 can't distinguish between `model` and `model-xyz`. Because of this, RV assumes that the `train` command doesn't need to be run again because model checkpoints
with a model prefix exist even though the actual model file isn't there.

### Notes

A more general solution that will avoid this problem in the future is for `report_io` to specify whether each input and output is a directory or a file, and then to set the `include_dir` argument to `file_exists` accordingly.

## Testing
Integration tests should tell if this works, and I also made sure old predict packages still work.

### DIFF
--- a/rastervision/backend/keras_classification/backend.py
+++ b/rastervision/backend/keras_classification/backend.py
@@ -131,7 +131,7 @@ class ModelFiles(FileGroup):
         """
         FileGroup.__init__(self, base_uri, tmp_dir)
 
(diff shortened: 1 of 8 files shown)


lewfish:
Closing this because there's a more general solution.

PR #773, opened 2019-04-26
Ensure valid geojson read from tippecanoe output.

lossyrob:
## Overview

On some systems, tippecanoe can output blank lines, which would cause an exception to be raised. This ensures the output is actually geojson by ensuring the line starts with `{`.


lewfish:
Tested that Spacenet Vegas still works using vector tiles.

lossyrob:
Since this isn't causing anyone else a bug besides my local laptop, I'm going to say this doesn't need a backport and can wait for the next release. We can re-evaluate if there are reports that this is a problem.

PR #776, opened 2019-05-06
Custom and Auxiliary Commands

lossyrob:
## Overview

This PR introduces custom commands. It allows users to write plugins that add commands to Raster Vision.

### Checklist

- [x] Implement
- [x] Test on AWS
- [x] Docs

### Notes

The following changes were needed for this feature:
- A `custom_config` property was added to ExperimentConfig. This is needed by the `get_root_uri` method of CommandConfig for custom commands; since there will not be a key or uri for custom commands (e.g. `chip_key` or `chip_uri`), there needs to be a place to set that information in the experiment, which the `custom_config` provides. It will be up to the custom command to check for what it needs and throw a `ConfigError` if the experiment doesn't have the necessary `custom_config` set.
- `rv.ALL_COMMANDS` moved to a method, `rv.all_commands()`. This will include any custom commands, appended to the list after the internal commands. Note that since the order of commands is determined by the input and output files determined by the command's `report_io`, we shouldn't have to worry about order here. The only time to worry about order is in the `CommandDefinition.from_experiments` method, where the resolution of configuration properties can have an ordering to them. Therefore, if you need one custom command to resolve via it's `upate_for_command` for one command before another, you will need to make sure the plugins are loaded in a specific order (i.e. register the commands in the order you want them to resolve).

This change came out of wanting to do some image pre-processing on AWS Batch, and this being the 2nd or 3rd time that I wish I had this feature so I could code it processing up as a plugin and parallelize over the images on Batch through Raster Vision.


docs/commands.rst
```diff
@@ -44,3 +45,177 @@ BUNDLE
...
+
+The ``COGIFY`` command will turn GDAL-readable images and turn them into `Cloud Optimized GeoTiffs <https://www.cogeo.org/>`_.
+
+See the CogifyCommand entry in the :ref:`aux command api` API docs for configuration options.
+
+
+Custom Commands
+---------------
+
+Custom Commands allow advanced Raster Vision users to implement their own commands using the :ref:`plugins` architecture.
```
lewfish: The docs leave me confused about the relationship between aux commands and custom commands. I think an aux command is just a type of custom command that was created using a helper that lets you leave out some of the details of a custom command. This might make more sense if you introduced custom commands first with an example. Then the next section could explain aux commands (as a special case of custom commands) and show how the previous example could be written more simply as an aux command.

lewfish: Also, I'm confused about how aux commands are configured vs custom commands. Is it the same for both?

docs/api.rst
```diff
@@ -370,7 +370,7 @@ rv.CHIP_CLASSIFICATION_EVALUATOR
 rv.OBJECT_DETECTION_EVALUATOR
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. autoclass:: rastervision.evaluation.ObjectDetectionEvaluatorConfigBuilder
+.. autoclass:: rastervisio.evaluation.ObjectDetectionEvaluatorConfigBuilder
```
lewfish: rastervisio -> rastervision

docs/commands.rst
```diff
@@ -44,3 +45,177 @@ BUNDLE
...
+------------------------
+
+Raster Vision utilizes *auxiliary commands* for things like data preperation. These are commands that do not run in the normal ML pipeline (e.g., if one were to run run ``rastervision run`` without an command specified). Auxiliary commands normally do not have the same type of implicit configuration setting as normal commands; because of this, file paths are often set explicity, and these commands are often configured and returned from an ``ExperimentSet`` method directly, instead of implicity created through the ``ExperimentConfig``.
+
+Configuring Aux Commands
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+There are two ways to configure an Aux command: one is through custom configuration set on an ``ExperimentConfig``, and the other is to directly return a ``CommandConfig`` instance from an experiment method. Normally Aux Commands are run separately from the normal experiment workflow, so we suggest returning command configurations as a default.
+
+Configurating an Aux Command from an ExperimentConfig
```
lewfish: Configurating -> Configuring

I think I noticed some other spelling errors in the docs updates so I would run them through a spell checker.

docs/commands.rst
```diff
@@ -44,3 +45,177 @@ BUNDLE
...
+Auxiliary (Aux) Commands
+------------------------
+
+Raster Vision utilizes *auxiliary commands* for things like data preperation. These are commands that do not run in the normal ML pipeline (e.g., if one were to run run ``rastervision run`` without an command specified). Auxiliary commands normally do not have the same type of implicit configuration setting as normal commands; because of this, file paths are often set explicity, and these commands are often configured and returned from an ``ExperimentSet`` method directly, instead of implicity created through the ``ExperimentConfig``.
+
+Configuring Aux Commands
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+There are two ways to configure an Aux command: one is through custom configuration set on an ``ExperimentConfig``, and the other is to directly return a ``CommandConfig`` instance from an experiment method. Normally Aux Commands are run separately from the normal experiment workflow, so we suggest returning command configurations as a default.
+
```
lewfish: If you return a command config from an experiment method, does this mean you cannot run any other commands for the experiment defined by that method? Is the reason you added support for the other option (setting the custom config on the `ExperimentConfig`) that you might want to run the aux command plus built-in command as part of a single Batch DAG?

lossyrob:
Review questions:

- How are (non-Aux) custom commands run by default? (Non-aux commands are always run by default)
- Explain better the difference between AuxCommands and CustomCommands.


Insert this https://whimsical.com/PFLa4RNCKoeykjP76rFqbZ

PR #811, opened 2019-08-06
Script to generate GeoTIFF from ZXY tile server

lewfish:
This PR adds a script to generate a GeoTIFF from a ZXY tile server URI schema, bounds in lng/lat format, and a zoom level. It correctly crops the image based on the bounds, and also works on TMS format tile servers. This lays the groundwork for a `ZXYRasterSource` and may be independently useful.

It is tested with unit tests, and I also ran it on an example. I created a tile set from a GeoTIFF using `gdal2tiles`, got some bounds in QGIS, ran the script, and checked that the output was correct. 

```
root@70d4aadebe15:/opt/src# python -m rastervision.data.raster_source.zxy2geotiff "/opt/data/hippo-tiles/{z}/{x}/{-y}.png" 17 \
>     "39.950380846608667 -75.163183288636375 39.951604868181825 -75.158675638636367" \
>     /opt/data/hippo-tiles/out.tif
```

in

out

### DIFF
--- a/rastervision/data/raster_source/zxy2geotiff.py
+++ b/rastervision/data/raster_source/zxy2geotiff.py
@@ -1,0 +1,182 @@
+import tempfile
+
+from PIL import Image
+import numpy as np
+import click
+import mercantile
+import rasterio
+from rasterio.windows import Window
+import pyproj
+
+from rastervision.utils.files import download_if_needed, get_local_path
+
+
+def lnglat2merc(lng, lat):
+    """Convert lng, lat point to x/y Web Mercator tuple."""
+    return pyproj.transform(
+        pyproj.Proj(init='epsg:4326'), pyproj.Proj(init='epsg:3857'), lng, lat)
+
+
+def merc2lnglat(x, y):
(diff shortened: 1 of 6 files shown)


lewfish:
Note: tiles in example were created using
```
conda install gdal
pip install gdal2tiles
gdal2tiles.py /opt/data/hippos.tif -z 17 /opt/data/hippo-tiles
```
in the fastai Docker image. (there was trouble installing it in the main RV image)

rastervision/data/raster_source/zxy2geotiff.py
```diff
@@ -0,0 +1,182 @@
+import tempfile
```
lossyrob: I understand why this is here (it'll be used by a raster source), but I think this would be more discoverable if it was moved to the `rastervision.utils` package. I can see these methods being imported as a utility by other code that just wants to grab geotiffs from zxy endpoints.

lewfish:
I updated according to your suggestion and added a `--make-cog` option using your code.

PR #812, opened 2019-08-13
Add ZXYRasterSource

lewfish:
This PR adds a `ZXYRasterSource` which allows reading imagery from zxy tile servers. It can be configured as follows:

```
        config = RasterSourceConfig.builder(rv.ZXY_RASTER_SOURCE) \
            .with_tile_schema('http://example.com/{z}/{x}/{y}.png') \
            .with_zoom(18) \
            .with_bounds([min_lat, min_lng, max_lat, max_lng]) \
            .build()
```

A limitation is that `ZXYRasterSources` can't be used with the `predict` command as it is currently implemented. This is because there is an assumption that a `RasterSource` can be instantiated using a single URI, but the `ZXYRasterSource` also requires a zoom level and bounds. So, currently this raises a `NotImplementedError`. This has been tested with unit tests, but still needs a real-world test as part of an experiment.


lewfish:
We decided it makes more sense to just use the zxy2geotiff script as a data preprocessing script, and not as a type of `RasterSource`. This will be more efficient (it will only need to be run once rather than every time the chip and predict command run as part of an experiment), and will also simplify the codebase.

PR #815, opened 2019-08-26
Bugfix chip command serialization of augmentors chip_command_config.py

lmbak:
Bugfix serialization of augmentors in the chip command

## Overview

It fixes a bug for the chip command where the augmentors were not properly serialized to the chip-command.json. By adding the augmentors objects to the to_proto method of the chip_command_config class this problem is resolved. I think this fix should also be backported.

### Checklist

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

Closes #XXX


lewfish:
LGTM. However, the CI job is failing because I recently made some changes to the way it's setup, but it's try to use the old setup for this PR. Could you please close the PR, change the branch name, and then make a new PR which the new branch name? I think that should fix it. Since this will get merged around the time of the new release, we don't need to worry about backporting it.

PR #816, opened 2019-08-28
Adds mirror augmentor using numpy matrices

lmbak:
## Overview

This adds a "mirror" augmentor that augments training data by adding a total of 11 copies (12 including the original) of the training data by mirroring along the x, y, and two diagonal axes. It is meant to be used for chip-classification, as the labels are not mirrored. I assume this would give invalid results for non-chip classification tasks. I have placed comments where the labels should be processed, but have not yet implemented the code. Not planning to do so on the short run. I have tested it with the chip classification on my own data; it works. It also means that the training data is increased by 12-fold.

### Checklist

### Notes

## Testing Instructions

* Testing can by done by creating a chip classification taks with the augmentor set on the dataset.

Closes #XXX


lmbak:
Should the pull requester finish all the checkboxes? Or is that part of your process of checking the pull request?

PR #828, opened 2019-09-20
Oversampling for the chip classification task

lmbak:
## Overview

This PR adds the option for the chip classification task to work with oversampling. This already exist for semantic segmentation, but I needed it also for the chip classification task. The oversampling for the segmentation works by class_id, while for the chip classification I chose to do it via the class_name. So now the user either inputs a list of class_name(s), OR one of class_ids. Oversampling via the id's for chip classification does not seem to be wise. See also the comment:

>         '''
>         For the chip classification the labels have two attributes:
>         "data", which is the index of the folder the image is in
>         and "obj", which is a string of the folder name the image is in.
>         Using the indices is not safe and can be confusing for the user
>         as the task.with_classes() method allows the user to specify
>         an id per class, which does not have to match the index of the folder.
>         Therefore it is safer to use the name of the folder instead of the index.
>         '''

### Checklist

## Testing Instructions

* I have tested this PR with the chip classification task. In that case the backend is specified as follows:
```
		backend = rv.BackendConfig.builder(rv.PYTORCH_CHIP_CLASSIFICATION) \
			.with_task(task) \
			.with_train_options(
				batch_size=20,
				weight_decay=1e-2,
				lr=1e-4,
				one_cycle=True,
				num_epochs=1,
				model_arch="resnet50",
				mixed_prec=True,
				flip_vert=True,
				sync_interval=1,
				debug=False,
				train_prop=1.0,
				train_count=None,
				log_tensorboard=True,
				run_tensorboard=True,
				oversample={'rare_class_name':['bridge'], 'rare_target_prop':0.5} \
				) \
			.build()
```


lmbak:
Closed after online-meeting with Rob en Lewis. Functionality will not be needed in the future, so pulling it now is not needed.

PR #856, opened 2019-11-14
Prioritize default temp dir

### LINKED ISSUE #855: Temp directory is not under /opt/data
The `RVConfig` object should be using `/opt/data` as the root of any temp dirs it creates by default. This is because that directory in the container is mounted to the host instance. Not doing using that directory will result in running out of disk space more quickly.

lewfish:
This PR prioritizes using the default temp dir location (/opt/data/tmp) over the system default location for temp dirs. By putting temp files in /opt/data which is mounted to the host instance when running on Batch, we can avoid running out of disk space when using large temporary files. 

I updated unit tests to cover this and also did a test run on Batch.

Closes #855


lossyrob:
Does this mean, in the local usage case, an `/opt/data/tmp` will be created by raster vision?

Would it make sense to check for the existence  of `/opt/data` before using this default? I think including some code that doesn't totally abstract over environment makes sense, because in the practical use case  we only have two (hoping that expands in the future...), but we also don't want to code too specifically to the AWS Batch use case, especially to a specific mounting strategy that we use.

lewfish:
> Does this mean, in the local usage case, an `/opt/data/tmp` will be created by raster vision?
> 
> Would it make sense to check for the existence of `/opt/data` before using this default? I think including some code that doesn't totally abstract over environment makes sense, because in the practical use case we only have two (hoping that expands in the future...), but we also don't want to code too specifically to the AWS Batch use case, especially to a specific mounting strategy that we use.

Yeah, it would create `/opt/data/tmp`. I'm cool with checking to see if `/opt/data` exists first.

lewfish:
I found that the `--tempdir` flag is passed to jobs that are submitted by the AWS Batch runner. I ran a regular command and the cogify command remotely, and in both cases the value set by the `--tempdir` flag (`/opt/data/temp`) was used correctly when downloading files from S3. So, I'm unable to replicate the problem. In any case, the changes in this PR don't seem like the right way to fix the problem, so I'm closing it.

PR #858, opened 2019-11-18
Oversampling for chip classification

### LINKED ISSUE #853: Oversampling of rare classes
Since the removal of Fastai backends there is no more option for oversampling (right? Or did I miss anything?). Would this be a usefull feature? In my case I would use it for the chip classification, where implementation might be easier than for semseg and object detection.

Would the correct construction be to provide the [DataLoader](https://github.com/azavea/raster-vision/blob/master/rastervision/backend/torch_utils/chip_classification/data.py#L24-L35) classes with a [Sampler ](https://pytorch.org/docs/stable/data.html#data-loading-order-and-sampler) object? More specifically the [WeightedRandomSampler](https://pytorch.org/docs/stable/data.html#torch.utils.data.WeightedRandomSampler)? The weights list for this sampler can then be calculated/constructed [somewhere in the DatasetFolder](https://github.com/azavea/raster-vision/blob/master/rastervision/backend/torch_utils/chip_classification/folder.py#L100-L116)?

@lewfish Could you let me know if we need this and if the described way is technically correct/how you would do it?

lmbak:
## Overview

This PR adds the ability to euqalize your class distribution through oversampling for the chip classification. ~~If you have three classes: A with 1001 images, B with 300 images, and C with 10 images a total of 1001*3=3003 images will be used for training. Each image in class B is used three times, with a remainder of 101 images chosen randomly from all the images in B. Each image in class C is used 100 times, with one image chosen at random. The images used are kept constant throughout training. It works by setting the samples in the ImageFolder to represent the distribution in which all classes are used euqally. The user can set the `equalize=True` flag in the `.with_train_options() ` of the `rv.BackendConfig.builder(rv.PYTORCH_CHIP_CLASSIFICATION)`~~

### Checklist

### Notes

## Testing Instructions

* Test on unequal distrubuted classes, result will be a training process on a set where the rare class gets sampled more.

Closes #853


lmbak:
Have not run `scripts/format_code` because this changes all the files.

rastervision/backend/torch_utils/chip_classification/data.py
```diff
@@ -21,6 +21,19 @@ def build_databunch(data_dir, img_sz, batch_sz, class_names):
         train_dir, transform=aug_transform, classes=class_names)
     valid_ds = ImageFolder(valid_dir, transform=transform, classes=class_names)
 
+    if equalize:
+        train_ds.samples = equalize_classes_through_oversampling(
```
lewfish: You should be able to implement this by setting the sampler on the `Dataloader` -- that's what it's there for. Modifying the `Dataset` doesn't seem like the right approach. There is an example of how to do this here: https://github.com/azavea/raster-vision-fastai-plugin/blob/a0283b0f25ae805efaf2abdef2ef35ec390b0746/fastai_plugin/semantic_segmentation_backend.py#L110

lmbak: Yes, you are right, that is what we discussed in the issue. I'll see to it that the right changes are made.

lmbak: @lewfish A question regarding that. I first wanted to try a [WeightedRandomSampler](https://pytorch.org/docs/stable/data.html#torch.utils.data.Sampler), but this could lead to the situation where you cannot be sure that each image is each epoch for training. So say that we have the images of class A and B: [A1,A2,A3,B1], and weight vector [1,1,1,3], we want to train with [A1, A2, A3, B1, B1, B1], but because of the WeightedRandomSampler we could end up with [A1,A2,A1,B1,B1,B1], meaning we miss image A3. Or would you say that this effect in the long run would be minimal and can be disregarded? Or is there a way in which this effect can be reached with the samplers (I could not figure it out)?

lewfish: Each epoch it will pick a different random set of indices, so I doubt it matters. But there is also the `replacement` argument to https://pytorch.org/docs/stable/data.html#torch.utils.data.WeightedRandomSampler

lmbak: Yes, only the replacement (when set to false) will not allow you to draw more images than there are in the dataset, so this would allow you to undersample, not oversample. When set to True the above situation can occur. But Maybe you're right and its too much of an edge-case to really matter. Maybe it might only matter with few epochs and little data, which is what we're trying to avoid anyway. I'll get working on it with the WeightedRandomSampler and the weights as per your example.

lmbak:
Still to do:
- manual testing (autmated local test seem to pass)
- remove oversampling for the validation dataset

I'll get working on this tomorrow, for now I'm calling it a day.

lmbak:
@lewfish manually test seems to work out. Also removed the `WeightedRandomSampler()` from the validation dataset. Again, if I run the `scripts/format_code` all files are changed. You just run the file from within the uppermost `raster-vision` folder, right?

lewfish:
> @lewfish manually test seems to work out. Also removed the `WeightedRandomSampler()` from the validation dataset. Again, if I run the `scripts/format_code` all files are changed. You just run the file from within the uppermost `raster-vision` folder, right?

I run `./scripts/format_code` from inside the container with working directory `/opt/src`.

lewfish:
Did you notice an improvement from using this functionality?

rastervision/backend/torch_utils/chip_classification/data.py
```diff
@@ -3,19 +3,49 @@
...
 
 from rastervision.backend.torch_utils.data import DataBunch
-from rastervision.backend.torch_utils.chip_classification.folder import (
-    ImageFolder)
+from rastervision.backend.torch_utils.chip_classification.folder import ImageFolder
 
 log = logging.getLogger(__name__)
 
 
+def calculate_oversampling_weights(imageFolder, rare_classes, desired_prob):
```
lewfish: Camel case shouldn't be used for variable or argument names in Python.

rastervision/backend/torch_utils/chip_classification/data.py
```diff
@@ -3,19 +3,49 @@
...
+    '''
+    Calculates weights tensor for oversampling
+
+    args:
+        imageFolder: instance of
```
lewfish: I would refactor this to take in the array of targets (ie. class ids) for the dataset, rather than the dataset itself. This way we could use it with any type of dataset.

rastervision/backend/torch_utils/chip_classification/data.py
```diff
@@ -3,19 +3,49 @@
...
+        desired prob: (float) a single probability that the rare classes should
+            have.
+    returns:
+        (tensor) with weights per index, e.g [0.5,0.1,0.9]
+    '''
+
+    chip_inds = []
+    for rare_class_id in rare_classes:
+        for idx, (sample, class_idx) in enumerate(imageFolder):
+            if class_idx == rare_class_id:
```
lewfish: You could remove the outer loop over `rare_classes` if you change this line to `if class_idx in rare_classes:`

rastervision/backend/torch_utils/chip_classification/data.py
```diff
@@ -74,13 +105,40 @@ def build_databunch(data_dir, img_sz, batch_sz, class_names, augmentors):
...
+            count_per_class = {}
+            for class_idx in imageFolder.class_to_idx.values():
+                count_per_class[class_idx] = 0
+            for (sample_path, class_index) in imageFolder.samples:
+                count_per_class[class_index] += 1
+            largest_class_idx = max(count_per_class, key=count_per_class.get)
+            return count_per_class[largest_class_idx]
+
+        num_train_samples = len(
+            train_ds.classes) * get_class_with_max_count(train_ds)
```
lewfish: I don't understand why you don't just use the `len(train_ds)` as the `num_train_samples`. The way it's set right now doesn't seem correct, but maybe I'm missing something.

lmbak:
> > @lewfish manually test seems to work out. Also removed the `WeightedRandomSampler()` from the validation dataset. Again, if I run the `scripts/format_code` all files are changed. You just run the file from within the uppermost `raster-vision` folder, right?
> 
> I run `./scripts/format_code` from inside the container with working directory `/opt/src`.

That did the trick, I used to run it from the raster-vision folder outside the container. When I do as you did it works (obviously)

rastervision/backend/torch_utils/chip_classification/data.py
lmbak: You're right, I missed this one.

rastervision/backend/torch_utils/chip_classification/data.py
lmbak: Done

rastervision/backend/torch_utils/chip_classification/data.py
lmbak: Done

rastervision/backend/torch_utils/chip_classification/data.py
lmbak: I sent you a long version explanation via mail, but the short version is that I will apply the method as you explained here

rastervision/backend/torch_utils/chip_classification/data.py
lmbak: Now set it to take in a list of class labels/targets. That should work for the chip classification. But in case of semseg or objdet i'm not sure that we can use the function in this form. Should the function only be usable for chipclassification?

rastervision/backend/torch_utils/chip_classification/data.py
```diff
@@ -74,13 +103,28 @@ def build_databunch(data_dir, img_sz, batch_sz, class_names, augmentors):
     valid_ds = AlbumentationDataset(
         ImageFolder(valid_dir, classes=class_names))
 
+    if rare_classes != []:
+        targets = [target[1] for target in train_ds.orig_dataset.imgs]
```
lewfish: It'd be more concise (and also make fewer assumptions about the private variables of `train_ds`) if you just used `[target for _, target in train_ds]`.

lmbak: Resolved as you proposed

lmbak:
I have squased a bit, the rest will be rather laborous because of conflicts. Is it currently acceptable or does is need to be cleaner? Because in that case I might create a clean branch and start over.

lmbak:
Considering the effort, i'll do that anyway. I'll close this one and send a 'clean' request.

PR #851, opened 2019-10-31
Add augmentors to chip classification

lmbak:
## Overview

Adds the option to augment training data via blurring, rotating, and/or adding snow to the training data. To do this the [albumentations ](https://github.com/albu/albumentations)library is used.

### Checklist

- [x] Updated `docs/changelog.rst`
- [x] Added `needs-backport` label if PR is bug fix that applies to previous minor release
- [ ] Ran scripts/format_code and committed any changes
- [x] Documentation updated if needed
- [x] PR has a name that won't get you publicly shamed for vagueness

### Notes

Still work in progress.

## Testing Instructions

* How to test this PR
* Prefer bulleted description
* Start after checking out this branch
* Include any setup required, such as rebuilding the Docker image.
* Include test case, and expected output if not captured by automated tests.

Closes #831

Issue #831, opened 2019-09-20
Add data augmentation to PyTorch backends


lewfish:
We would like to add an option for training data augmentation to the new PyTorch backends in PR #830 . A good start would be adding it to the chip classification backend. This is roughly where the code should go: https://github.com/azavea/raster-vision/blob/c1594b211ed3cd267b451248e08a50a70f35cc1c/rastervision/backend/torch_utils/chip_classification/data.py#L16-L20

Note this link might not work once this PR is merged into master. 

We should use the albumentations library https://github.com/albu/albumentations. Here is an example of using it for object detection from another codebase (which formed the basis of the object detection backend in RV): https://github.com/lewfish/mlx/blob/88e21d9fdd2f0d7b3effb750a65b64e308bccc08/mlx/od/data.py#L154-L173

lewfish:
@lmbak

lmbak:
Work in progress at https://github.com/lmbak/raster-vision/tree/lmb/augmentation

lmbak:
@lewfish I am currently testing, but run into the following error:

```
Traceback (most recent call last):
  File "/opt/conda/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/opt/conda/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/opt/src/rastervision/__main__.py", line 17, in <module>
    rv.main()
  File "/opt/conda/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/opt/conda/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/opt/conda/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/conda/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/conda/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/opt/src/rastervision/cli/main.py", line 294, in run_command
    rv.runner.CommandRunner.run(command_config_uri)
  File "/opt/src/rastervision/runner/command_runner.py", line 11, in run
    CommandRunner.run_from_proto(msg)
[...]

PR #859, opened 2019-11-18
Add augmentors to chip classification

lewfish:
This PR adds augmentation options for PyTorch chip classification using the albumentations library.

I tested this by using the `ToGray` augmentation on a sample chip classification experiment and saw that some of the debug chips were greyscale.

This supercedes #851 

Connects #831


lewfish:
@lmbak I made some changes to get this over the line faster. The main change is that I created an `AlbumentationsDataset` adapter which makes it so we don't need to make any changes to the original `Dataset`.

lewfish:
Also, I made it so the validation dataset doesn't have any data augmentations applied to it. This is a best practice that reduces the variance in the validation metrics from one run to the next.

lmbak:
Great! Thanks for the feedback. In line with your comment I'm assuming that oversampling is also not applied to the validation set, I'll fix that in the corresponding PR.

PR #915, opened 2020-04-29
Provide GDAL VSI filesystem plugin

### LINKED ISSUE #905: Add VSI file system
If we added a `FileSystem` plugin that handled any GDAL VSI uri we might never need to write another file system again. See https://gdal.org/user/virtual_file_systems.html

jpolchlo:
## Overview

Files may be sourced from a wide array of endpoints, and piecemeal implementation of those different target schemas would take time.  GDAL provides a virtual file system interface that provides support for many basic schemes (s3, Google cloud storage, http, ftp, and file), with the ability to read (and sometimes write) files from inside archives stored on those remote targets.

Supported schema are s3, http, https, ftp, gs, and file, which may be combined with the archive schema (tar, zip, gzip).  For example,
```
 zip:tar:s3://jpolchlopek/test.tar/!test/test.zip!test.tif
```
reads a TIF from inside a ZIP, which is stored in a TAR on S3.  These URIs may be formed as
```
<archive_scheme>:<archive_URI>!path/to/contained/file.ext
```
(Note: we may need to use a `+` in place of the `:` for Rasterio compatibility, though the colon notation seems to be in line with [JAR access](https://docs.oracle.com/javase/6/docs/api/java/net/JarURLConnection.html).)

### Checklist

### Notes

This PR requires at least version 2.3 of GDAL.  Current docker images supply 2.2.2.  I have developed against GDAL 3.0.4.

Also, attempted to run `scripts/format_code`, and the set of files affected was very large, so for the sake of PR hygiene, I reverted.

## Testing Instructions (TBD)

* How to test this PR
* Prefer bulleted description

Closes #905

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,14 +6,10 @@ RUN apt-get update && apt-get install -y software-properties-common python-softw
 RUN add-apt-repository ppa:ubuntugis/ppa && \
     apt-get update && \
     apt-get install -y wget=1.* git=1:2.* python-protobuf=2.* python3-tk=3.* \
-                       gdal-bin=2.2.* \
                        jq=1.5* \
                        build-essential libsqlite3-dev=3.11.* zlib1g-dev=1:1.2.* \
                        libopencv-dev=2.4.* python-opencv=2.4.* unzip curl && \
     apt-get autoremove && apt-get autoclean && apt-get clean
-
-# Setup GDAL_DATA directory, rasterio needs it.
-ENV GDAL_DATA=/usr/share/gdal/2.2/
 
 # See https://github.com/mapbox/rasterio/issues/1289
 ENV CURL_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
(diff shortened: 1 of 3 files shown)


jamesmcclain:
> zip:tar:s3://jpolchlopek/test.tar/!test/test.zip!test.tif

I would be in favor (at least optionally) supporting the normal GDAL style of describing paths `/vsizip/vsitar/vsis3/...`.

jamesmcclain:
> This PR requires at least version 2.3 of GDAL.

Is this for feature-completeness, or to function at all?

lewfish:
> Also, attempted to run `scripts/format_code`, and the set of files affected was very large, so for the sake of PR hygiene, I reverted.

Yeah, I usually put those changes in a separate commit.

lewfish:
Overall this looks great.

> This PR requires at least version 2.3 of GDAL. Current docker images supply 2.2.2. I have developed against GDAL 3.0.4.

Feel free to upgrade it in the Dockerfile.

jpolchlo:
> > This PR requires at least version 2.3 of GDAL.
> 
> Is this for feature-completeness, or to function at all?

While investigating the VSI capabilities of GDAL, I ran across some references that suggested that ... something I was doing required 2.3.  I didn't make a note of what it was.  I'll make a comment if I can dig up the reference again.

> > zip:tar:s3://jpolchlopek/test.tar/!test/test.zip!test.tif
> 
> I would be in favor (at least optionally) supporting the normal GDAL style of describing paths `/vsizip/vsitar/vsis3/...`.

After chatting, I think I'll move to supporting VSI paths directly, while retaining the function to convert URI-structured paths to VSI paths.

> > This PR requires at least version 2.3 of GDAL. Current docker images supply 2.2.2. I have developed against GDAL 3.0.4.
> 
> Feel free to upgrade it in the Dockerfile.

I've done this, but I wonder if `gdal==3.0.4` should be included in the `requirements.txt` or one of its friends, even though the Conda install of gdal includes the correct python package?

lewfish:
> I've done this, but I wonder if `gdal==3.0.4` should be included in the `requirements.txt` or one of its friends, even though the Conda install of gdal includes the correct python package?

I think that would be useful to have if someone was installing RV using pip, so I would try adding it. But if it breaks something then don't worry about it. I'm going to be doing a refactor/audit of the whole Docker and pip configuration before the release, so if something is tricky hopefully I can get it to work then.

PR #917, opened 2020-05-01
Fix Mapbox style-based vector rasterization

jpolchlo:
## Overview

When a `class_id` is not provided in the `properties` field of a GeoJSON, we may need to specify a test to determine what class is applicable.  This is done using a dictionary of the following form:
```python
filter_dict = {
    0: ['==', 'wacky_field', 'background'],
    1: ['!=', 'wacky_field', 'background']
}
```

This was not functioning previously; this PR aims to fix that.

### Checklist

## Testing Instructions

* How to test this PR
* Prefer bulleted description

Closes #XXX

### DIFF
--- a/rastervision2/core/data/vector_source/vector_source_config.py
+++ b/rastervision2/core/data/vector_source/vector_source_config.py
@@ -4,6 +4,7 @@ from rastervision2.pipeline.config import Config, register_config, Field
 
 ClassFilter = Sequence[Union[str, 'ClassFilter']]
 
+
 @register_config('vector_source')
 class VectorSourceConfig(Config):
     default_class_id: Optional[int] = Field(
--- a/rastervision2/core/data/vector_source/vector_source_config.py
+++ b/rastervision2/core/data/vector_source/vector_source_config.py
@@ -1,4 +1,4 @@
-from typing import Dict, Optional, Union
+from typing import Dict, Optional, Union, Sequence
 
 from rastervision2.pipeline.config import Config, register_config, Field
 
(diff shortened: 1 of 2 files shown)


rastervision2/core/data/vector_source/vector_source_config.py
```diff
@@ -11,7 +11,7 @@ class VectorSourceConfig(Config):
         ('The default class_id to use if class cannot be inferred using other '
          'mechanisms. If a feature defaults to a class_id of None, then that feature '
          'will be deleted.'))
-    class_id_to_filter: Optional[Dict[int, Optional[Dict]]] = Field(
```
lewfish, line 14: I think this can contain nested lists so it should just be `Sequence`, not `Sequence[str]`

jpolchlo: I looked into this, and it doesn't appear to work using a nested sequence.  It makes some sense that it wouldn't since it's not clear if multiple tests would be processed as an 'any of' or 'all of' relationship.  I'm having trouble finding examples of Mapbox styles that illustrate how this would work, so any pointers would be useful.

Should we proceed with single tests or actually update the code to do the right thing with lists of lists (assuming an _all of_ approach for lists of qualifications)?

lewfish: I found the docs for the filter language: https://docs.mapbox.com/mapbox-gl-js/style-spec/other/#other-filter (please update the description string while you're at it) From this it's apparent that lists of lists are possible. I don't think we need to update anything in the RV code for this to work. This is where it's implemented: https://github.com/azavea/raster-vision/blob/master/rastervision/data/vector_source/label_maker/filter.py

jpolchlo: Aha.  Makes sense.  Fixed now.

rastervision2/core/data/vector_source/vector_source_config.py
```diff
@@ -1,7 +1,8 @@
-from typing import Dict, Optional, Union
+from typing import Dict, Optional, Sequence, Union
 
 from rastervision2.pipeline.config import Config, register_config, Field
 
+ClassFilter = Sequence[Union[str, 'ClassFilter']]
```
lewfish, line 5: Nice.

PR #923, opened 2020-06-01
Convert CRS to a type interpretable by pyproj

jpolchlo:
## Overview

Rasterio datasets produce their CRS responses in `rasterio.crs.CRS`.  These cannot be passed directly to the `pyproj.Proj` constructor.  Converting to WKT prior to use was sufficient to fix the bug.

### Checklist

- [ ] Updated `docs/changelog.rst`
- [ ] Added `needs-backport` label if PR is bug fix that applies to previous minor release
- [ ] Ran scripts/format_code and committed any changes
- [x] Documentation updated if needed
- [x] PR has a name that won't get you publicly shamed for vagueness

### Notes

This may be an artifact of versions, as I am using a development container that relies on GDAL 3 and may have a more recent Rasterio version installed, causing problems.

PR #926, opened 2020-06-05
Batch runner naming and configuration fixes

jpolchlo:
## Overview

There is an unfortunate bit of naming in the RV2 runner for batch jobs that, when combined with Everett, makes for a bit of irritation in use.  We have the `aws_batch` runner, which has a config section in `~/.rastervision/default` headed by `[AWS_BATCH]`.  This generally causes no problems until one encounters a case where environment variables are easier to use than a config file.  The case I encountered was launching an RV2 pipeline from a batch process, in preparation for creating Granary tasks which do the same.  (For purposes of discussion, I'll say there is a "base" batch job which is triggered by Granary and that calls `rastervision2`, initiating "pipeline" batch jobs which do the work.)

I can set the environment variables like `AWS_BATCH_GPU_JOB_QUEUE` in the base job definition, but upon execution, these environment variables do not appear.  According to [this page](https://docs.aws.amazon.com/batch/latest/userguide/job_env_vars.html):
> All variables that are set by AWS Batch begin with the prefix, AWS_BATCH_. This is a protected environment variable prefix, and you cannot use this prefix for your own variables in job definitions or overrides. 

As such, it is impossible to configure the `aws_batch` runner via environment variable.

This PR renames `aws_batch` to `batch_runner` (I'll take suggestions for better names), and after doing so, the deployment pattern that will be used by Granary (base batch job kicking off pipeline jobs) works as expected.

The renaming needed to make this PR complete isn't entirely done, but I wanted to put up the incomplete version for discussion.

### Notes

This is built on top of #923 .  It doesn't need to be; I've done this for my own testing purposes.  I'll need to rebase before a merge, if the PR is accepted.

### Checklist


lewfish:
I would prefer renaming it to `batch` rather than `batch_runner` which is inconsistent with the other runners which don't have `_runner` in the name.

PR #940, opened 2020-06-17
Handle different versions of pyproj

jpolchlo:
## Overview

When using pipelines in derived environments that have later versions of pyproj installed, a problem arises with RasterioCRSTransformer failing to correctly use the CRS from a Rasterio data source.  This PR introduces a version check and provides correct logic for either case.

Supersedes #923 

### Checklist

### Notes

The requirement for versions of pyproj above 2.2.0 is imposed by s2cloudless, used by cloud-buster, so this seems like a necessary feature to handle.

## Testing

In a container, install s2cloudless, and run `python -m unittest discover tests` and `tests_v2`.  Both test commands work before and after the s2cloudless install.

### DIFF
--- a/rastervision/data/crs_transformer/rasterio_crs_transformer.py
+++ b/rastervision/data/crs_transformer/rasterio_crs_transformer.py
@@ -16,14 +16,10 @@ class RasterioCRSTransformer(CRSTransformer):
             image_dataset: Rasterio DatasetReader
             map_crs: CRS code
         """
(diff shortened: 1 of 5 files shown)


lewfish:
Why not just upgrade pyproj in this repo, and then we can avoid the the version check?

jpolchlo:
Wasn't sure there was a reason for picking the pyproj version in the container.  Upgraded the installed version to 2.6.1.post1.  If there's a preferred version, let me know.

Issue #566, opened 2018-11-15
Integration test for backward compatibility


lewfish:

lewfish:
Can we use 0.8 predict packages with develop?

lewfish:
I think we're switching to there being no backward compatibility for model bundles, which would make this a moot issue. Instead model bundles will be linked to Docker image tags that they can be run within. But I'm not totally sure yet, so leaving this open.

PR #972, opened 2020-08-13
Add support for multiband images for semantic segmentation

### LINKED ISSUE #961: Multiband support for semantic segmentation
This issue is for adding support for training models on > 3 bands for semantic segmentation. The main idea is to save and use images in numpy array format instead of PNG format. Adding support for other tasks, and adding automated testing and documentation is outside the scope of this issue.

## Testing Strategy
* First, make sure you can run the ISPRS Potsdam example since we will base our test off of it. See https://docs.rastervision.io/en/latest/examples.html#semantic-segmentation-isprs-potsdam and https://github.com/azavea/raster-vision/blob/master/rastervision_pytorch_backend/rastervision/pytorch_backend/examples/semantic_segmentation/isprs_potsdam.py
* Modify the example so that it has an extra `multiband` argument [here](https://github.com/azavea/raster-vision/blob/master/rastervision_pytorch_backend/rastervision/pytorch_backend/examples/semantic_segmentation/isprs_potsdam.py#L15)
* If the `multiband` option is True, then use all four bands as the channel order [here](https://github.com/azavea/raster-vision/blob/master/rastervision_pytorch_backend/rastervision/pytorch_backend/examples/semantic_segmentation/isprs_potsdam.py#L59).
* When developing this, you will want to run it locally, and using `-a test True`. I also find it very helpful to use the debugger. I use the debugger in Visual Studio Code, and have some support for this using `docker/run --debug` and inside the container `./scripts/debug rastervision.pipeline.cli run ...`. You will need to configure the remote debugger in VS Code using something like the following to do volume mounting:
```
        {
            "name": "Raster Vision: Remote Attach",
            "type": "python",
            "request": "attach",
            "port": 3003,
            "host": "localhost",
            "pathMappings": [
                {
                    "localRoot": "/Users/lfishgold/projects/raster-vision",
                    "remoteRoot": "/opt/src"
                }
            ],
            "justMyCode": false
        },
```
* Publish image to ECR using `docker/ecr_publish`, and run the experiment on Batch.
* Check that ndarray files are actually produced in the `chip` directory on S3. The scores in the `eval` directory should be better than or equal to running with only three bands.

## Implementation Strategy
* In `rastervision.pytorch_backend`, write chips in [`ndarray`](https://numpy.org/doc/stable/reference/generated/numpy.save.html) format in https://github.com/azavea/raster-vision/blob/master/rastervision_pytorch_backend/rastervision/pytorch_backend/pytorch_semantic_segmentation.py#L15 We should use ndarrays of a type that matches the resolution of the GeoTIFF in order to save space. For example, if the GeoTIFF is uint16, then we should use an ndarray with `dtype` equal to `np.uint16`.
* In `rastervision.pytorch_learner`,  make dataset read chips in ndarray format. We should leave the existing ability to read PNG files intact. (This is because the `rastervision.pytorch_learner` package is meant to be usable outside of a geospatial context, and PNGs are a more standard way of distributing training data.) We can do this by adding a new data format in https://github.com/azavea/raster-vision/blob/master/rastervision_pytorch_learner/rastervision/pytorch_learner/semantic_segmentation_learner_config.py#L10 and add a new `Dataset` that handles `ndarray` datasets that is similar to https://github.com/azavea/raster-vision/blob/master/rastervision_pytorch_learner/rastervision/pytorch_learner/semantic_segmentation_learner.py#L23. Alternatively, we could just modify the `SemanticSegmentationDataset` to handle both png and ndarray files. That could be simpler, but I'm not sure.
  * A potential issue is that `albumentations` (the data augmentation library) might not be able to handle > 3 bands (I haven't looked into it).
  * Another issues is that dividing by 255.0 to get a number between 0 and 1 which the model expects (https://github.com/azavea/raster-vision/blob/master/rastervision_pytorch_learner/rastervision/pytorch_learner/semantic_segmentation_learner.py#L42) will no longer be correct if the ndarrays have more resolution than uint8.
*  To build a model that can handle > 3 bands, we will need to modify the DeeplabV3 model before it is returned in the `build_model` method: https://github.com/azavea/raster-vision/blob/master/rastervision_pytorch_learner/rastervision/pytorch_learner/semantic_segmentation_learner.py#L52 @jamesmcclain has code for doing this. Handling the `pretrained` option correctly will require some work because the pretrained model only natively handles 3 bands.
- The `plot_xyz` method needs to be modified so it picks three of the bands for display purposes: https://github.com/azavea/raster-vision/blob/master/rastervision_pytorch_learner/rastervision/pytorch_learner/semantic_segmentation_learner.py#L132 To select the bands to use, a new option can be added to `SemanticSegmentationDataConfig` https://github.com/azavea/raster-vision/blob/master/rastervision_pytorch_learner/rastervision/pytorch_learner/semantic_segmentation_learner_config.py#L15
* Note that any configuration options added to `rastervision.pytorch_learner` will need to also be added to https://github.com/azavea/raster-vision/blob/master/rastervision_pytorch_backend/rastervision/pytorch_backend/pytorch_semantic_segmentation_config.py#L14 and then forwarded similar to https://github.com/azavea/raster-vision/blob/master/rastervision_pytorch_backend/rastervision/pytorch_backend/pytorch_semantic_segmentation_config.py#L21

AdeelH:
## Overview

- Adds support for multiband images for the semantic segmentation task.
  - When using more than 3 bands, creates/stores/consumes training/validation/prediction chips in the form of numpy arrays (`.npy`).
  - Automatically modifies the Deeplab model to accommodate the additional channels
    - If not using pretrained weights, the first conv layer is replaced with one with the correct number of channels
    - If using pretrained weights, a new conv layer for the new channels is added in parallel to the original first conv layer, and their parallel outputs are summed, before passing on to the rest of the network
- The evals of a model trained with all 4 RGBIR bands was about the same as an RGB one with the same hyperparameters
  - Outputs of the runs are available at
    - RGB: s3://raster-vision-ahassan/potsdam/rgb/output
    - RGBIR: s3://raster-vision-ahassan/potsdam/data/output
- Improved plots of the input image, ground truth labels, and predicted labels.
    - Option to customize the plotting of individual bands of the image. For example, if using an input with bands [R, G, B, IR, Elevation], it is now possible to have it plotted as 3 images - one for RGB, one for IR, and one for Elevation.
    - Sample output: s3://raster-vision-ahassan/potsdam/data/output/train/test_preds.png

### Checklist

### Notes

- `SemanticSegmentationDataset` is now able to handle both `.npy` and `.png` inputs
- New options have been added to `class SemanticSegmentationDataConfig(DataConfig)`
  - `img_channels`, `channel_display_groups`, `img_format`, `label_format`
- The same options have also been added to `class SemanticSegmentationConfig(RVPipelineConfig)` so they can be passed to the above
- `.update()` for the above configs initializes `channel_display_groups` based on `img_channels` if not specified by the user

## Testing Instructions

* How to test this PR
  - Specify the `channel_order` in the isprs_potsdam example
  - Pass `-a multiband True` to rastervision when running the above example

Closes #961

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,7 +7,7 @@ Raster Vision 0.13
 Features
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-* Added support for multiband images `#961 <https://github.com/azavea/raster-vision/pull/961>`_
+* Added support for multiband images `#972 <https://github.com/azavea/raster-vision/pull/972>`_
 
 
 Raster Vision 0.12
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,9 +7,7 @@ Raster Vision 0.13
 Features
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-* It is now possible to use images with more than 3 bands for the semantic segmentation task. Raster Vision automatically modifies the PyTorch DeepLab model to accommodate the additional channels. If using a pretrained model, the pretrained weights for the RGB bands are retained. When using more than 3 bands, Raster Vision creates/consumes training/validation/prediction chips in the form of numpy arrays.
-* Improved plots of the input image, ground truth labels, and predicted labels.
-* Option to customize the plotting of individual bands of the image. For example, if using an input with bands [R, G, B, IR, Elevation], it is now possible to have it plotted as 3 images - one for RGB, one for IR, and one for Elevation.
+* Added support for multiband images `#961 <https://github.com/azavea/raster-vision/pull/961>`_
 
 
 Raster Vision 0.12
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,17 @@
 CHANGELOG
 =========
 
+Raster Vision 0.13
+-------------------
+
+Features
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* It is now possible to use images with more than 3 bands for the semantic segmentation task. Raster Vision automatically modifies the PyTorch DeepLab model to accommodate the additional channels. If using a pretrained model, the pretrained weights for the RGB bands are retained. When using more than 3 bands, Raster Vision creates/consumes training/validation/prediction chips in the form of numpy arrays.
+* Improved plots of the input image, ground truth labels, and predicted labels.
+* Option to customize the plotting of individual bands of the image. For example, if using an input with bands [R, G, B, IR, Elevation], it is now possible to have it plotted as 3 images - one for RGB, one for IR, and one for Elevation.
+
+
 Raster Vision 0.12
 -------------------
 
--- a/rastervision_core/rastervision/core/data/dataset_config.py
+++ b/rastervision_core/rastervision/core/data/dataset_config.py
@@ -32,8 +32,9 @@ class DatasetConfig(Config):
             for s in self.test_scenes:
                 s.update(pipeline=pipeline)
 
-        self.img_channels = len(
-            self.train_scenes[0].raster_source.channel_order)
+        channel_order = self.train_scenes[0].raster_source.channel_order
+        if channel_order is not None:
+            self.img_channels = len(channel_order)
 
     def validate_config(self):
         ids = [s.id for s in self.train_scenes]
--- a/rastervision_core/rastervision/core/data/dataset_config.py
+++ b/rastervision_core/rastervision/core/data/dataset_config.py
@@ -1,9 +1,12 @@
-from typing import List
+from typing import List, Optional
+from pydantic import PositiveInt
 
-from rastervision.pipeline.config import Config, register_config, ConfigError
+from rastervision.pipeline.config import (Config, register_config, ConfigError,
+                                          Field)
 from rastervision.pipeline.utils import split_into_groups
 from rastervision.core.data.scene_config import SceneConfig
 from rastervision.core.data.class_config import ClassConfig
+from rastervision.core.data.utils import all_equal
 
 
 @register_config('dataset')
@@ -13,6 +16,9 @@ class DatasetConfig(Config):
     train_scenes: List[SceneConfig]
     validation_scenes: List[SceneConfig]
     test_scenes: List[SceneConfig] = []
+
+    img_channels: Optional[PositiveInt] = Field(
+        None, description='The number of channels of the images.')
 
     def update(self, pipeline=None):
         super().update()
@@ -26,6 +32,9 @@ class DatasetConfig(Config):
             for s in self.test_scenes:
                 s.update(pipeline=pipeline)
 
+        self.img_channels = len(
+            self.train_scenes[0].raster_source.channel_order)
+
     def validate_config(self):
         ids = [s.id for s in self.train_scenes]
         if len(set(ids)) != len(ids):
@@ -35,6 +44,14 @@ class DatasetConfig(Config):
         if len(set(ids)) != len(ids):
             raise ConfigError(
                 'All validation and test scene ids must be unique.')
+
+        self.ensure_same_channel_order()
+
+    def ensure_same_channel_order(self):
+        all_scenes = self.train_scenes + self.validation_scenes + self.test_scenes
+        channel_orders = [s.raster_source.channel_order for s in all_scenes]
+        if not all_equal(channel_orders):
+            raise ConfigError('channel_order must be same for all scenes.')
 
     def get_split_config(self, split_ind, num_splits):
         new_cfg = self.copy()
--- a/rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation_config.py
+++ b/rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation_config.py
@@ -88,16 +88,45 @@ class SemanticSegmentationConfig(RVPipelineConfig):
             img_channels = min(3, self.dataset.img_channels)
             self.channel_display_groups = {'Input': tuple(range(img_channels))}
 
+    def validate_config(self):
+        super().validate_config()
+        self.validate_channel_display_groups()
+
     def get_default_label_store(self, scene):
         return SemanticSegmentationLabelStoreConfig()
 
     def get_default_evaluator(self):
         return SemanticSegmentationEvaluatorConfig()
 
-    def update(self):
-        super().update()
+    def validate_channel_display_groups(self):
+        def _are_ints(ints) -> bool:
+            return all(isinstance(i, int) for i in ints)
 
-        self.dataset.class_config.ensure_null_class()
+        def _in_range(inds, lt: int) -> bool:
+            return all(0 <= i < lt for i in inds)
 
-        if self.channel_display_groups is None:
-            self.channel_display_groups = [tuple(range(self.img_channels))]
+        img_channels = self.dataset.img_channels
+        groups = self.channel_display_groups
+
+        # validate dict form
+        if isinstance(groups, dict):
+            for k, v in groups.items():
+                if not isinstance(k, str):
+                    raise ConfigError(
+                        'channel_display_groups keys must be strings.')
+                if not isinstance(v, (list, tuple)):
+                    raise ConfigError(
+                        'channel_display_groups values must be lists or tuples.'
+                    )
+                if not (_are_ints(v) and _in_range(v, lt=img_channels)):
+                    raise ConfigError(
+                        f'Invalid channel indices in channel_display_groups[{k}].'
+                    )
+
+        # validate list/tuple form
+        if isinstance(groups, (list, tuple)):
+            for i, grp in enumerate(groups):
+                if not (_are_ints(grp) and _in_range(grp, lt=img_channels)):
+                    raise ConfigError(
+                        f'Invalid channel index in channel_display_groups[{i}].'
+                    )
--- a/rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation_config.py
+++ b/rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation_config.py
@@ -89,6 +89,9 @@ class SemanticSegmentationConfig(RVPipelineConfig):
 
         self.dataset.class_config.ensure_null_class()
 
+        if self.dataset.img_channels is None:
+            return
+
         if self.img_format is None:
             self.img_format = 'png' if self.dataset.img_channels == 3 else 'npy'
 
--- a/rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation_config.py
+++ b/rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation_config.py
@@ -99,6 +99,9 @@ class SemanticSegmentationConfig(RVPipelineConfig):
     def validate_config(self):
         super().validate_config()
 
+        if self.dataset.img_channels is None:
+            return
+
         if self.img_format == 'png' and self.dataset.img_channels != 3:
             raise ConfigError('img_channels must be 3 if img_format is png.')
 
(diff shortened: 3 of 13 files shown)


lewfish:
Did you notice any accuracy improvement when using the multiband option? That would be a good way to test if everything's working.

lewfish:
> Did you notice any accuracy improvement when using the multiband option? That would be a good way to test if everything's working.

Nevermind, I didn't see your message above that said you did this.

jamesmcclain:
> If not using pretrained weights, the first conv layer is replaced with one with the correct number of channels

> If using pretrained weights, a new conv layer for the new channels is added in parallel to the original first conv layer, and their parallel outputs are summed, before passing on to the rest of the network

Does this present portability concerns?  Raster-vision does not package architecture code into its model bundles, so some provision will have to be made to remember which kind of model is in a bundle (one created from pretrained weights or one created from random weights, because the architectures will be different).

AdeelH:
>Does this present portability concerns? Raster-vision does not package architecture code into its model bundles, so some provision will have to be made to remember which kind of model is in a bundle (one created from pretrained weights or one created from random weights, because the architectures will be different).

That's a very good point and one that has not been addressed in this PR.

I will need look into how to serialize model architectures in PyTorch. This might be a good starting point: https://stackoverflow.com/questions/59287728/saving-pytorch-model-with-no-access-to-your-code

AdeelH:
To clarify, it is a portability concern if using the model outside of Raster Vision. Within RV, there should be no problem, because it uses the same model for eval.

jamesmcclain:
> To clarify, it is a portability concern if using the model outside of Raster Vision. Within RV, there should be no problem, because it uses the same model for eval.

Okay, as long as raster-vision knows what architecture to use when handed an well-formed model bundle, I think we are in good shape.  (I thinking of both the `eval` command and the `predict` command, here.)

@lewfish may which to provide input.

lewfish:
> > To clarify, it is a portability concern if using the model outside of Raster Vision. Within RV, there should be no problem, because it uses the same model for eval.
> 
> Okay, as long as raster-vision knows what architecture to use when handed an well-formed model bundle, I think we are in good shape. (I am thinking of both the `eval` command and the `predict` command, here.)
> 
> I would like (eventually) to see the architecture included in the model bundle if for no other reasons than archivability and forward-compatibility.
> 
> @lewfish may which to provide input.

I agree that the problem is solved for the sake of this PR, but a separate issue should be made for archiving the model architecture for the sake of using it outside of RV. I'm not very knowledgeable about TorchScript, but that seems like a good option.

lewfish:
I made a separate issue re: archival models: https://github.com/azavea/raster-vision/issues/974

rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation_config.py
```diff
@@ -61,6 +61,17 @@ class SemanticSegmentationConfig(RVPipelineConfig):
     chip_options: SemanticSegmentationChipOptions = SemanticSegmentationChipOptions(
     )
 
+    img_channels: int = Field(
```
lewfish, line 64: It seems like `img_channels` and `img_format` could and should be determined from the `channel_order` in `RasterSourceConfig`. If `len(channel_order) == 3`, then set the format to png. If > 3, then use npy. This will help users avoid invalid configuration. I think it makes sense to still have the ability to manually override things, but the default should be to avoid needing extra configuration. 

More concretely, I'm suggesting that we make the `img_channels` and `img_format` Optional with default value None, and then make the `update` method fill in those values based on the length of the `channel_order` array in the `RasterSourceConfig` in the first scene.

Also, it would be good to validate `img_channels` and `img_format` so that they are consistent with one another. For instance, `img_channels=4` and `img_format='png'` would be invalid.

rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation_config.py
```diff
@@ -61,6 +61,17 @@ class SemanticSegmentationConfig(RVPipelineConfig):
...
+        3, description='The number of channels of the training images.')
+
+    channel_display_groups: Optional[Union[dict, list, tuple]] = Field(
```
lewfish, line 67: A little more information in the description would be good since the docs are generated from this. Specifically what the content of the dict or list or tuple should contain.

rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation_config.py
AdeelH: That makes sense. I have changed it so that `DatasetConfig` sets a `img_channels` property on itself based on the first training scene. And since this assumes that `channel_order` for all scenes must be the same, I have added a validation check for that.

`img_format` is also now initialized based on `img_channels`.

rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation_config.py
AdeelH: Done. It now reads:
>Groups of image channels to display together as a subplot when plotting the data and predictions. Can be a list or tuple of groups (e.g. [(0, 1, 2), (3,)]) or a dict containing title-to-group mappings (e.g. {"RGB": [0, 1, 2], "IR": [3]}), where each group is a list or tuple of channel indices and title is a string that will be used as the title of the subplot for that group.

AdeelH:
>Aside from the two comments, remember to update the change log.

Should I add this under v0.12 in changelog.rst?

lewfish:
> Should I add this under v0.12 in changelog.rst?

No, I would put it in a new section for 0.13, since 0.12 is the latest released version.

rastervision_core/rastervision/core/data/dataset_config.py
```diff
@@ -14,6 +17,9 @@ class DatasetConfig(Config):
     validation_scenes: List[SceneConfig]
     test_scenes: List[SceneConfig] = []
 
+    img_channels: Optional[PositiveInt] = Field(
+        None, description='The number of channels of the images.')
```
lewfish, line 21: This broke backward compatibility with a model bundle from the 0.12 release. Anytime we add a new option with a *required value* to a config, or change the name of an option, we need to use a config upgrader to maintain backward compatibility for model bundles. Or, if it seems reasonable, you can add a default value and you don't need an upgrader.

You can see an example of adding a config version upgrader here: https://github.com/azavea/raster-vision/commit/9806a48dbeaea9972130eaeb1702f831f6a83f39 If you need more info on how the config upgrader works, we can have a quick meeting.

To test this kind of thing, I've been manually running the predict command with models from the Model Zoo https://docs.rastervision.io/en/latest/examples.html#model-zoo But, there is an issue to add this to the integration tests which we might want to prioritize now. https://github.com/azavea/raster-vision/issues/566

```
root@85d5380e10ec:/opt/src# rastervision predict https://s3.amazonaws.com/azavea-research-public-data/raster-vision/examples/model-zoo-0.12/isprs-potsdam-ss/model-bundle.zip \
>      https://s3.amazonaws.com/azavea-research-public-data/raster-vision/examples/model-zoo-0.12/isprs-potsdam-ss/3_12_sample.tif \
>      /opt/data/examples/spacenet-vegas-roads-ss/out.tif
Traceback (most recent call last):
  File "/opt/conda/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/opt/conda/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/opt/src/rastervision_pipeline/rastervision/pipeline/cli.py", line 248, in <module>
    main()
  File "/opt/conda/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/opt/conda/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/opt/conda/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/conda/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/conda/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/opt/src/rastervision_core/rastervision/core/cli.py", line 66, in predict
    predictor.predict([image_uri], output_uri)
  File "/opt/src/rastervision_core/rastervision/core/predictor.py", line 101, in predict
    self.pipeline.predict()
  File "/opt/src/rastervision_core/rastervision/core/rv_pipeline/rv_pipeline.py", line 149, in predict
    self.backend = self.config.backend.build(self.config, self.tmp_dir)
  File "/opt/src/rastervision_pytorch_backend/rastervision/pytorch_backend/pytorch_semantic_segmentation_config.py", line 39, in build
    learner = self.get_learner_config(pipeline)
  File "/opt/src/rastervision_pytorch_backend/rastervision/pytorch_backend/pytorch_semantic_segmentation_config.py", line 25, in get_learner_config
    channel_display_groups=pipeline.channel_display_groups)
  File "pydantic/main.py", line 274, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for SemanticSegmentationDataConfig
img_channels
  none is not an allowed value (type=type_error.none.not_allowed)
```

docs/changelog.rst
```diff
@@ -1,6 +1,17 @@
 CHANGELOG
 =========
 
+Raster Vision 0.13
+-------------------
+
+Features
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* It is now possible to use images with more than 3 bands for the semantic segmentation task. Raster Vision automatically modifies the PyTorch DeepLab model to accommodate the additional channels. If using a pretrained model, the pretrained weights for the RGB bands are retained. When using more than 3 bands, Raster Vision creates/consumes training/validation/prediction chips in the form of numpy arrays.
```
lewfish, line 10: For features with associated PRs we just write the name of the PR and a link to it. See example here: https://github.com/azavea/raster-vision/blob/master/docs/changelog.rst#raster-vision-011

There were more verbose notes for the 0.12 release which was an anomaly because it was a big refactoring that was spread across many PRs.

rastervision_core/rastervision/core/data/dataset_config.py
lewfish: A little more info: the `img_channels` field was added to `SemanticSegmentationDataConfig` and is required. But old model bundles don't supply that field since it didn't exist. So the config upgrader needs to upgrade the version from 0 to 1, and set a value of 3 for `img_channels`. Each plugin has its own config version (which is currently 0 for `pytorch_backend`) so this will need to be incremented in `pytorch_backend/__init__.py`

rastervision_core/rastervision/core/data/dataset_config.py
```diff
@@ -26,6 +32,9 @@ def update(self, pipeline=None):
             for s in self.test_scenes:
                 s.update(pipeline=pipeline)
 
+        self.img_channels = len(
+            self.train_scenes[0].raster_source.channel_order)
```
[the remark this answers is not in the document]
AdeelH: If we throw an error on `channel_order` being `None`, it will effectively make it non-optional, which I don't think is desirable. And in the absence of `channel_order`, there is also no way to determine it from the config alone, without building a scene first. 

I have therefore modified `SemanticSegmentation(RVPipeline)` so that if `DatasetConfig.img_channels` is not set, it builds the first scene to extract `.raster_source.num_channels` and sets `DatasetConfig.img_channels` to it. It then also calls `.update()` and `.validate()` on it for good measure.

I think this is a good solution that does not disturb anything else.

rastervision_core/rastervision/core/data/dataset_config.py
AdeelH: I have set a default value of 3 instead of using an upgrader. It seems reasonable because prior to this PR, RV only used 3-channel images.

I tried testing on the model bundles from the model zoo, but they seem to have an incompatible config structure. In particular, they don't have a learner config.

docs/changelog.rst
AdeelH: Got it

rastervision_core/rastervision/core/data/dataset_config.py
lewfish: I just tested using a bundle from the model zoo, and it worked after the changes you made. 

```
rastervision predict https://s3.amazonaws.com/azavea-research-public-data/raster-vision/examples/model-zoo-0.12/isprs-potsdam-ss/model-bundle.zip \
     https://s3.amazonaws.com/azavea-research-public-data/raster-vision/examples/model-zoo-0.12/isprs-potsdam-ss/3_12_sample.tif \
     /opt/data/examples/isprs-potsdam/out.tif
```

PR #977, opened 2020-08-20
Add support for weighted loss for classification and semantic segmentation

### LINKED ISSUE #867: Use a weighted loss function instead of oversampling
The oversampling functionality allows imbalanced classes to be shown equally during the training process. However, some form of randomness is included because images are randomly sampled. In an extreme case, or a case with few epochs, this would mean that not all images are used during training, which is undesirable as that would mean a loss of information shown to the model.

One way to solve this could be to use a weighted loss function. In the [Documentation](https://pytorch.org/docs/stable/nn.html) of the CrossEntropyLoss function it is explained that a weights vector can be passed to the loss function. This way the randomness would be removed from the training process, likely leading to more reproducible results.

AdeelH:
## Overview

- Adds support for using a class weighted loss for classification and semantic segmentation (#867) 
    - custom weights can be provided as the optional parameter, `class_loss_weights`, to `SolverConfig`
        - must be provided as `list` or `tuple` since pydantic does not support serialization of numpy arrays and torch Tensors (if weights are generated as numpy arrays and torch Tensors, they can be conveniently converted to `list` using `.tolist()`)
    - if provided, the weights are converted to a `torch.Tensor` and passed as the `weight` parameter to `nn.CrossEntropyLoss`
    - validation for `class_loss_weights` involves checking if its length is equal to `len(DataConfig.class_names)`

### Checklist

### Notes

- Also fixes a bug in a call to `GeoJSONVectorSourceConfig` in the chip classification example
- Not added to object detection, because the classification loss is calculated [deep within the FasterRCNN model](https://github.com/pytorch/vision/blob/master/torchvision/models/detection/roi_heads.py#L35) and it does not expose an API to provide weights for it.

## Testing Instructions

* How to test this PR
    - e.g. pass `class_loss_weights=torch.ones(len(class_config.names) + 1).tolist()` to SolverConfig in `isprs_potsdam.py`
    - e.g. pass `class_loss_weights=torch.ones(len(class_config.names)).tolist()` to SolverConfig in `isprs_potsdam.py`
    - on passing all zeros, the loss should be zero

Closes #867

### DIFF
--- a/rastervision_pytorch_backend/rastervision/pytorch_backend/pytorch_learner_backend_config.py
+++ b/rastervision_pytorch_backend/rastervision/pytorch_backend/pytorch_learner_backend_config.py
@@ -1,7 +1,10 @@
 from typing import List
 
-from rastervision.pipeline.config import register_config, Field
+from rastervision.pipeline.config import (register_config, Field, validator,
+                                          ConfigError)
 from rastervision.core.backend import BackendConfig
+from rastervision.pytorch_backend import (PyTorchSemanticSegmentationConfig,
+                                          PyTorchChipClassificationConfig)
 from rastervision.pytorch_learner.learner_config import (
     SolverConfig, ModelConfig, default_augmentors, augmentors as
     augmentor_list)
@@ -36,3 +39,12 @@ class PyTorchLearnerBackendConfig(BackendConfig):
 
     def build(self, pipeline, tmp_dir):
         raise NotImplementedError()
+
+    @validator('solver')
+    def validate_solver_config(cls, v):
+        if v.class_loss_weights is not None:
+            if cls not in (PyTorchSemanticSegmentationConfig,
+                           PyTorchChipClassificationConfig):
+                raise ConfigError(
+                    'class_loss_weights is currently only supported for '
+                    'Semantic Segmentation and Chip Classification.')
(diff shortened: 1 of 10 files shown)


rastervision_pytorch_learner/rastervision/pytorch_learner/learner_config.py
```diff
@@ -150,6 +150,8 @@ class SolverConfig(Config):
          'epochs with start and end LR being lr/10 and the peak being lr.'))
     multi_stage: List = Field(
         [], description=('List of epoch indices at which to divide LR by 10.'))
+    class_loss_weights: Optional[Union[list, tuple]] = Field(
```
lewfish, line 153: If someone sets this field for a task that is not semantic segmentation or chip classification, a validation error should be raised.

rastervision_pytorch_backend/rastervision/pytorch_backend/pytorch_learner_backend_config.py
```diff
@@ -36,3 +37,16 @@ def get_learner_config(self, pipeline):
...
+        if v.class_loss_weights is not None:
+            from rastervision.pytorch_backend import (
+                PyTorchSemanticSegmentationConfig,
+                PyTorchChipClassificationConfig)
+            if cls not in (PyTorchSemanticSegmentationConfig,
+                           PyTorchChipClassificationConfig):
+                raise ConfigError(
+                    'class_loss_weights is currently only supported for '
+                    'Semantic Segmentation and Chip Classification.')
+        return v
```
AdeelH, line 52: @lewfish, I added a check here to ensure it's only used for SS and CC.

PR #992, opened 2020-09-11
Add support for external loss function definition

### LINKED ISSUE #983: Add support for external loss functions
It should be possible to use `torch.hub` to load external loss functions the same way that it is used to load external architectures (#985). This will allow the use of arbitrary loss functions, which is currently not possible due to the requirement that the configs be serializable.

@lewfish has suggested that instead of adding focal loss to RV (#982), we can have it in a separate repo and use it through this mechanism where needed.

AdeelH:
## Overview

Adds support for loading the loss function via torch.hub. Currently not supported for object detection.

- An external loss definition can be specified by supplying an `ExternalModuleConfig` to `SolverConfig.external_loss_def`.
- Example usage:
	- classification (`spacenet_rio.py`)
	```python3
    if external_loss:
        external_loss_def = ExternalModuleConfig(
            github_repo='AdeelH/pytorch-multi-class-focal-loss',
            name='focal_loss',
            entrypoint='focal_loss',
            force_reload=False,
            entrypoint_kwargs={
                'alpha': [.75, .25],
                'gamma': 2
            })
    else:
        external_loss_def = None

    solver = SolverConfig(
        lr=1e-2,
        num_epochs=20,
        test_num_epochs=4,
        batch_sz=32,
        one_cycle=True,
        external_loss_def=external_loss_def)
	}))
	```
- Example usage included in `spacenet_rio.py`, and can be enabled through the `external_loss` argument.
- When loading from a bundle, the definition is recovered and its path provided to `learner.__init__()` via a new `loss_def_path` param.
- Everything else works the same way as in #985.

### Checklist

## Testing Instructions

* How to test this PR
    - Use example above. Try with different options.

Closes #983

### DIFF
--- a/rastervision_pytorch_backend/rastervision/pytorch_backend/examples/chip_classification/spacenet_rio.py
+++ b/rastervision_pytorch_backend/rastervision/pytorch_backend/examples/chip_classification/spacenet_rio.py
@@ -111,7 +111,7 @@ def get_config(runner,
         external_loss_def = None
 
     solver = SolverConfig(
-        lr=1e-2,
+        lr=1e-4,
         num_epochs=20,
         test_num_epochs=4,
         batch_sz=32,
--- a/rastervision_pytorch_backend/rastervision/pytorch_backend/examples/chip_classification/spacenet_rio.py
+++ b/rastervision_pytorch_backend/rastervision/pytorch_backend/examples/chip_classification/spacenet_rio.py
@@ -19,7 +19,8 @@ def get_config(runner,
                processed_uri,
                root_uri,
                test=False,
-               external_model=False):
+               external_model=False,
+               external_loss=False):
     debug = False
     train_scene_info = get_scene_info(join(processed_uri, 'train-scenes.csv'))
     val_scene_info = get_scene_info(join(processed_uri, 'val-scenes.csv'))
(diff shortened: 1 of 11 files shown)


rastervision_pytorch_backend/rastervision/pytorch_backend/examples/chip_classification/spacenet_rio.py
```diff
@@ -96,8 +97,27 @@ def make_scene(scene_info):
...
+            entrypoint_kwargs={
+                'alpha': [.75, .25],
+                'gamma': 2
+            })
+    else:
+        external_loss_def = None
+
     solver = SolverConfig(
-        lr=1e-4, num_epochs=20, test_num_epochs=3, batch_sz=32, one_cycle=True)
+        lr=1e-2,
```
lewfish, line 114: Did you mean to change the LR? I would have thought that LR would not work.

rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
```diff
@@ -211,37 +215,75 @@ def setup_model(self, model_def_path: str = None) -> None:
...
         pass
 
+    def setup_loss(self, loss_def_path: Optional[str] = None) -> None:
+        """Setup self.loss.
+
+        Args:
+            loss_def_path (str, optional): Loss definition path. Will be
+            available when loading from a bundle. Defaults to None.
+        """
+        extCfg = self.cfg.solver.external_loss_def
```
lewfish, line 238: Is there a reason why `extCfg` isn't named `ext_cfg` (following the usual Python naming conventions)?

lewfish (changes requested):

rastervision_pytorch_backend/rastervision/pytorch_backend/examples/chip_classification/spacenet_rio.py
AdeelH: Nope. Didn't mean to commit this.

rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
AdeelH: That's a good point. I have no idea why I didn't notice that before.

lewfish (approved):

Issue #1003, opened 2020-09-22
Multi-Raster Source is Hard-Coded Not to Use Transformers (In Some Places)


jamesmcclain:
While trying to use the multi-raster source with input rasters of different type (float32 and float64), I had to write a raster transformer  to make all of the sub-raster sources the same type (which is completely expected).  The unexpected part is that `use_transformers` is hard-coded to `False` in at least one place, which causes a failure in the `analyze` stage (because the transformer is not able to run and change the `dtype` of the float64 raster).

In my case the line where that occurs is [here](https://github.com/azavea/raster-vision/blob/a2f7a1b52ed1a63053c3571771699fefe6e7ed10/rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation.py#L101).

AdeelH:
I think it would break at both the analyze stage 
https://github.com/azavea/raster-vision/blob/a2f7a1b52ed1a63053c3571771699fefe6e7ed10/rastervision_core/rastervision/core/rv_pipeline/rv_pipeline.py#L50

and  `SemanticSegmentation.get_image_channels()` which uses code copied from the above
https://github.com/azavea/raster-vision/blob/a2f7a1b52ed1a63053c3571771699fefe6e7ed10/rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation.py#L100-L101

The simple solution is to just change them to `True`. I don't believe that would effect MultiRasterSource in any way. Not sure about the `analyze` stage.

PR #1004, opened 2020-09-23
Remove Some Assumptions from MultiRasterSource

jamesmcclain:
Optionally remove dtype and size assumptions.

In lieu of https://github.com/azavea/raster-vision/issues/1003

## Overview

### Checklist

### Notes

## Testing Instructions

Run the `analyze` and/or `chip` commands on a project that uses the MultiRasterSource and has sub-rasters of different dtypes and/or (slightly) different extents.

### DIFF
--- a/rastervision_core/rastervision/core/data/raster_source/multi_raster_source.py
+++ b/rastervision_core/rastervision/core/data/raster_source/multi_raster_source.py
@@ -1,4 +1,4 @@
-from typing import Optional, Sequence, Union
+from typing import Optional, Sequence
 from pydantic import conint
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


AdeelH:
Why are we doing it this way instead of using transformers and enabling `use_transformers` everywhere?

rastervision_core/rastervision/core/data/raster_source/multi_raster_source.py
```diff
@@ -19,18 +19,31 @@ class MultiRasterSource(ActivateMixin, RasterSource):
...
             raster_sources (Sequence[RasterSource]): Sequence of RasterSources.
             raw_channel_order (Sequence[conint(ge=0)]): Channel ordering that
                 will always be applied before channel_order.
+            force_subchip_size_fill_value (Optional[Union[int, float]]):
+                Value used to pad sub-chips so that they are all the same size as the
+                sub-chip from the first sub-source.  This is to accommodate small
+                differences in sub-raster size such those that might be caused by
+                attempting to reproject disparate sub-rasters into the same projection.
+                No special reprojection logic is triggered by this parameter.  Use with
+                caution.
```
AdeelH, line 43: Isn't the output of `RasterSource._get_chip()` supposed to be the same size as specified by the window it takes in?

jamesmcclain: Well, the size of the overall chip ends up being the same size as the first subchip of the multiraster source.

In most instances, this will have no effect at all.  The idea here is to handle cases where there are multiple rasters which cover the same area but (in my case) differ by 1 in their width.  When that case is encountered, the extra column will be filled with the fill value in subchip from the smaller raster.

jamesmcclain:
> Why are we doing it this way instead of using transformers and enabling `use_transformers` everywhere?

Good question.  There are two reasons:

The first is that I might eventually need raster transformers for all possible base types.  When running entirely locally those can be implemented in the pipeline, but if one wishes operate on Batch, custom images or upstream inclusion of those transformers is required.  This behavior seems to be simpler and more parsimonious than the full set of transformers.

The second reason is that I realized that I need behavior to handle slight differences in raster size due to reprojection, and that seemingly can only be handled at this level.  Since (I think) I need to make a change here, it made sense to make the other change that I need, as well.

rastervision_core/rastervision/core/data/raster_source/multi_raster_source.py
AdeelH: Looking at the code for `RasterioSource`, it seems like `window` _is_ supposed to be in pixel coordinates and `RasterioSource` returns a chip of the height and width specified by the `window`.

Disabling the validation check for equal extents makes sense though.

AdeelH:
> I might eventually need raster transformers for all possible base types

If by base types you mean dtypes, then a general `toDtype` transformer might be simpler. Alternatively, we could also make use of the existing `RasterSource.dtype` (or add a `target_dtype` property), and change `RasterSource` so that it casts to this `dtype` before returning chips. This will make each `RasterSource` responsible for its return type.

jamesmcclain:
> > I might eventually need raster transformers for all possible base types
> 
> If by base types you mean dtypes, then a general `toDtype` transformer might be simpler. Alternatively, we could also make use of the existing `RasterSource.dtype` (or add a `target_dtype` property), and change `RasterSource` so that it casts to this `dtype` before returning chips. This will make each `RasterSource` responsible for its return type.

These sound like reasonable suggestions.

lewfish:
These changes look ok to me.

lewfish:
Adeel's approach may be more generic or elegant, but since this is already done, and there's a good chance MultiRasterSource will only be used on one project, I'm inclined to accept the change. We can revisit it later if it's too limited for what we need.

AdeelH (approved):

jamesmcclain:
@lewfish Good to merge?  (I will try wait until after the CI issues are resolved.)

lewfish (approved):

PR #1007, opened 2020-09-24
Give Cache A Chance

jamesmcclain:
## Overview

Hold VSI handles open (rather than closing them immediately) in the GDAL VSI file system to allow GDAL's caching behavior to operate.

### Checklist

### Notes

## Testing Instructions

The particular use-case that motivated this PR was one in which a small number of very large vector label files were shared by a larger number of scenes.  In that instance, the re-downloaded of the vector labels was quite noticeable.

To test, use the GDAL VSI file system to repeatedly read assets of some kind.  If the environment variables `VSI_CACHE` and `VSI_CACHE_SIZE` are appropriately set, one should be able to detect a difference.


jamesmcclain:
This seems not to be needed.  Just changing the URI so that GDAL is used seems to be enough ...

PR #1008, opened 2020-09-25
Make num_workers Specifiable

jamesmcclain:
## Overview

The requirement for shared memory comes from the fact that worker/reader processes used for chip reading use shared memory to send chips back to the main process.  Setting this to a number lower than the default 4 reduces pressure, and setting it to 0 removes the need for shared memory entirely.

### Checklist

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

Closes #XXX

### DIFF
--- a/rastervision_pytorch_backend/rastervision/pytorch_backend/pytorch_semantic_segmentation_config.py
+++ b/rastervision_pytorch_backend/rastervision/pytorch_backend/pytorch_semantic_segmentation_config.py
@@ -21,6 +21,7 @@ class PyTorchSemanticSegmentationConfig(PyTorchLearnerBackendConfig):
             img_channels=pipeline.dataset.img_channels,
             img_format=pipeline.img_format,
             label_format=pipeline.label_format,
+            num_workers=self.num_workers,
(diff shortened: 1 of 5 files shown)


rastervision_pytorch_backend/rastervision/pytorch_backend/pytorch_semantic_segmentation_config.py
```diff
@@ -21,6 +21,7 @@ def get_learner_config(self, pipeline):
             img_channels=pipeline.dataset.img_channels,
             img_format=pipeline.img_format,
             label_format=pipeline.label_format,
+            num_workers=self.num_workers,
```
lewfish, line 24: I think this change should be made to the other learner configs (eg.  in pytorch_object_detection_config.py). Otherwise, it would be set by the user, but then wouldn't be used for other tasks.

lewfish (changes requested):

lewfish:
This also needs a changelog update.

lewfish (approved):

PR #1025, opened 2020-10-08
Allow filtering out chips based on proportion of NODATA pixels

### LINKED ISSUE #999: Filter out images that are >x% NODATA
Currently, RV does not filter out chips that are mostly NODATA and therefore not useful for training/validation. Example chip:

2fdf89e12a049f33ac20cafdf561c930c93ca211d072e7121846cf7f8cde20d4-1852

Would be useful to have an option to filter these chips out of the dataset.

AdeelH:
## Overview

- Add `chip_nodata_threshold` field to `RVPipelineConfig`. 
- Filter out chips where the proportion of NODATA values in the chip is greater than or equal to this value.

### Checklist

### Notes

- Unrelated change: simplified code for `fill_no_data()` in `core/rv_pipeline/semantic_segmentation.py`.

Closes #999

### DIFF
--- a/rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation.py
+++ b/rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation.py
@@ -11,8 +11,11 @@ from rastervision.core.rv_pipeline.semantic_segmentation_config import (
 log = logging.getLogger(__name__)
 
 
-def get_train_windows(scene, class_config, chip_size,
-                      chip_options) -> List[Box]:
+def get_train_windows(scene,
+                      class_config,
+                      chip_size,
+                      chip_options,
+                      chip_nodata_threshold=1.) -> List[Box]:
     """Get training windows covering a scene.
 
     Args:
@@ -33,9 +36,8 @@ def get_train_windows(scene, class_config, chip_size,
             filt_windows = []
             for w in windows:
                 label_arr = label_source.get_labels(w).get_label_arr(w)
-                null_inds = (
-                    label_arr.ravel() == class_config.get_null_class_id())
-                if not np.all(null_inds):
+                null_mask = label_arr == class_config.get_null_class_id()
+                if null_mask.mean() < chip_nodata_threshold:
                     filt_windows.append(w)
             windows = filt_windows
         return windows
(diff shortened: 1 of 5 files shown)


rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation.py
```diff
@@ -81,14 +83,10 @@ def should_use_window(window):
...
-    null_inds = label_arr.ravel() == null_class_id
-    img_shape = img.shape
-    if np.any(null_inds):
-        img = np.reshape(img, (-1, img_shape[2]))
-        img[null_inds, :] = 0
-        img = np.reshape(img, img_shape)
+    # If chip has null labels, fill in those pixels with nodata.
+    mask = label_arr == null_class_id
+    if np.count_nonzero(mask) > 0:
+        img[mask] = 0
```
lewfish, line 89: Nice, I didn't know you could do this.

rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation.py
```diff
@@ -33,9 +36,8 @@ def filter_windows(windows):
             filt_windows = []
             for w in windows:
                 label_arr = label_source.get_labels(w).get_label_arr(w)
-                null_inds = (
-                    label_arr.ravel() == class_config.get_null_class_id())
-                if not np.all(null_inds):
+                null_mask = label_arr == class_config.get_null_class_id()
```
lewfish, line 39: Shouldn't we be filtering based on whether or not the image from the raster source has NODATA pixels, rather than based on null classes in the label array? (In other words, shouldn't you do the same thing that you did in object detection and chip classification?

lewfish (changes requested):

rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation.py
AdeelH: It was already filtering based on null labels and not based on NODATA pixels like the other pipelines, so I assumed those were equivalent. But I guess that is not necessarily true.

lewfish (approved):

PR #1027, opened 2020-10-10
Allow ignore_last_class w/ External Loss Functions

jamesmcclain:
## Overview

Allow the last class to be ignored when an external loss function is used.  This is done by passing `'force'` to `ignore_last_class` rather than `True` or `False`.

### Checklist

### DIFF
--- a/rastervision_pytorch_learner/rastervision/pytorch_learner/learner_config.py
+++ b/rastervision_pytorch_learner/rastervision/pytorch_learner/learner_config.py
@@ -2,6 +2,7 @@ from os.path import join
 from enum import Enum
 
 from typing import (List, Optional, Union, TYPE_CHECKING)
+from typing_extensions import Literal
 from pydantic import PositiveFloat, PositiveInt, constr
 
 from rastervision.pipeline.config import (Config, register_config, ConfigError,
@@ -195,7 +196,7 @@ class SolverConfig(Config):
         [], description=('List of epoch indices at which to divide LR by 10.'))
     class_loss_weights: Optional[Union[list, tuple]] = Field(
         None, description=('Class weights for weighted loss.'))
-    ignore_last_class: bool = Field(
+    ignore_last_class: Union[bool, Literal['force']] = Field(
         False,
         description=('Whether to ignore the last class during training.'))
     external_loss_def: Optional[ExternalModuleConfig] = Field(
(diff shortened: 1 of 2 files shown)


AdeelH:
If I am understanding this and #1017 correctly, `ignore_last_class=True` does 2 things:
	1. sets `ignore_idx` in `CrossEntropyLoss()`
	2. modifies the model so that it predicts one class less

In case of an external loss function, (1) does not happen, but one might still want (2), though it is not strictly necessary (because if the loss func is ignoring that index anyway, modifying the model does not really have much effect. An unmodified model would not receive any gradient on the weights that predict that last class anyway).

So, I think if all this is clearly noted in the in the documentation for `ignore_last_class`, `'force'` is not needed and just 'True' will suffice.

jamesmcclain:
> If I am understanding this and #1017 correctly, `ignore_last_class=True` does 2 things:
> 1. sets `ignore_idx` in `CrossEntropyLoss()`
> 2. modifies the model so that it predicts one class less
> 
> In case of an external loss function, (1) does not happen, but one might still want (2), though it is not strictly necessary (because if the loss func is ignoring that index anyway, modifying the model does not really have much effect. An unmodified model would not receive any gradient on the weights that predict that last class anyway).
> 
> So, I think if all this is clearly noted in the in the documentation for `ignore_last_class`, `'force'` is not needed and just `True` will suffice.

It is true that the last class does not get any attention from the gradient, but that is part of the problem.  It is clear that its existence still has some effect on the model because I noticed the `null` class continuing to be predicted well into training when starting from a pretrained model.

AdeelH:
Fair enough. I still think the description should be clearer about what the difference between `True` and `'force'` is.

rastervision_pytorch_learner/rastervision/pytorch_learner/learner_config.py
```diff
@@ -210,7 +211,7 @@ def validate_config(self):
         has_weights = self.class_loss_weights is not None
         has_external_loss_def = self.external_loss_def is not None
 
-        if self.ignore_last_class and has_external_loss_def:
+        if self.ignore_last_class is True and has_external_loss_def:
             raise ConfigError(
                 'ignore_last_class is not supported with external_loss_def.')
```
AdeelH, line 216: And maybe this error should tell the user to use `'force'` instead

AdeelH (changes requested):

jamesmcclain:
@AdeelH I have addressed your comment

AdeelH (approved):

lewfish (approved):

PR #1028, opened 2020-10-11
Only Normalize Unsigned Int Types in Prediction

jamesmcclain:
## Overview

Normalization is only performed for unsigned int types in the prediction stage.

### Checklist

### DIFF
--- a/rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
+++ b/rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
@@ -664,7 +664,11 @@ class Learner(ABC):
             the same tensor that has been scaled to [0-1].
 
         """
-        return x.float() / 255.0
+        if x.dtype == np.uint8 or x.dtype == np.uint16 or x.dtype == np.uint32:
+            limit = np.iinfo(x.dtype).max
+            return x.float() / float(limit)
+        else:
+            return x.float()
 
     def predict(self,
                 x: Tensor,
@@ -716,8 +720,7 @@ class Learner(ABC):
         x = torch.tensor(x)
         x = self.to_batch(x)
         x = x.permute((0, 3, 1, 2))
-        out = self.predict(
-            x, normalize=self.cfg.predict_normalize, raw_out=raw_out)
+        out = self.predict(x, normalize=True, raw_out=raw_out)
         return self.output_to_numpy(out)
 
     def predict_dataloader(self,
--- a/rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
+++ b/rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
@@ -654,41 +654,21 @@ class Learner(ABC):
             x = x.unsqueeze(0)
         return x
 
-    def normalize_input(self, x: Tensor) -> Tensor:
-        """Normalize an input image to have values between 0 and 1.
-
-        Args:
-            x: an image or batch of images assumed to be in uint8 format
-
-        Returns:
-            the same tensor that has been scaled to [0-1].
-
-        """
-        if x.dtype == torch.uint8:
-            limit = float(np.iinfo(np.uint8).max)
-        else:
-            limit = 1.0
-
-        return x.float() / limit
-
     def predict(self,
                 x: Tensor,
-                normalize: bool = False,
                 raw_out: bool = False) -> Any:
         """Make prediction for an image or batch of images.
 
         Args:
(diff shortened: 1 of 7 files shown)


rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
```diff
@@ -715,7 +715,8 @@ def numpy_predict(self, x: np.ndarray, raw_out: bool = False) -> Any:
         x = torch.tensor(x)
         x = self.to_batch(x)
         x = x.permute((0, 3, 1, 2))
-        out = self.predict(x, normalize=True, raw_out=raw_out)
+        out = self.predict(
```
jamesmcclain, line 718: It might make sense to cast `x` to type float (float32) in `predict` when normalization is not done because it seems that `x` must be the same type as the weights.

lewfish: Sounds good

lewfish:
Why / when would a model not be trained with normalization? Is it just if the model was trained outside of RV? The semantic segmentation DataSet is doing normalization based on the dtype of the imagery here https://github.com/azavea/raster-vision/blob/master/rastervision_pytorch_learner/rastervision/pytorch_learner/semantic_segmentation_learner.py#L82-L84 It seem like we should just normalize the same way at prediction time as at training time, but that doesn't seem to be the case. Am I missing something @AdeelH? I'm not opposed to adding this option, but it seems like a better solution in most cases is to do normalization the same way for prediction.

AdeelH:
> It seem like we should just normalize the same way at prediction time as at training time

I agree. It makes sense to do that in `Learner.numpy_predict()`. I think `Learner.predict()` should not be responsible for normalization at all and should expect its input to be in the right format already.

jamesmcclain:
> Why / when would a model not be trained with normalization? Is it just if the model was trained outside of RV? The semantic segmentation DataSet is doing normalization based on the dtype of the imagery here https://github.com/azavea/raster-vision/blob/master/rastervision_pytorch_learner/rastervision/pytorch_learner/semantic_segmentation_learner.py#L82-L84 It seem like we should just normalize the same way at prediction time as at training time, but that doesn't seem to be the case. Am I missing something @AdeelH? I'm not opposed to adding this option, but it seems like a better solution in most cases is to do normalization the same way for prediction.

The motivation for this is that I noticed normalization was not being done when I was working on `float32` rasters.  I am open to any solution that allows the `predict` stage to give me correct answers.

jamesmcclain:
@lewfish This PR has been updated in accordance with our conversation.

lewfish (approved):

lewfish (changes requested):
Oops, I'm cancelling that approval. For once I think the integration tests are telling us something useful. I'm guessing that the input was already converted to floats before passing it into the normalize function which is causing it to bypass the normalization which is necessary.

lewfish:
I'm looking into how to fix this now...

rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
```diff
@@ -664,7 +664,11 @@ def normalize_input(self, x: Tensor) -> Tensor:
             the same tensor that has been scaled to [0-1].
 
         """
-        return x.float() / 255.0
+        if x.dtype == np.uint8 or x.dtype == np.uint16 or x.dtype == np.uint32:
```
lewfish, line 667: x is a torch tensor so you need to use torch.uint8, etc or the equality check won't work.

lewfish (changes requested):

rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
jamesmcclain: Hmmm ... it seems as though `torch.uint16` and `torch.uint32` do not exist.

AdeelH:
@jamesmcclain I made a PR to this branch https://github.com/jamesmcclain/raster-vision/pull/2 uses the same type conversion code as `SemanticSegmentationDataset` and does the conversion while the input is still a numpy array.

lewfish:
> @jamesmcclain I made a PR to this branch ([jamesmcclain#2](https://github.com/jamesmcclain/raster-vision/pull/2)) that uses the same type conversion code as `SemanticSegmentationDataset` and does the conversion while the input is still a numpy array.

That looks more sound.

lewfish (approved):

PR #1032, opened 2020-10-16
Type cast before saving chips

AdeelH:
## Overview

This is a workaround for a problem encountered in the WB project where the tifs were uint16, but almost all the pixel values were < 256, with the rest few being larger.

### Checklist

### DIFF
--- a/rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation_config.py
+++ b/rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation_config.py
@@ -78,12 +78,13 @@ class SemanticSegmentationConfig(RVPipelineConfig):
         None, description='The filetype of the training images.')
     label_format: str = Field(
         'png', description='The filetype of the training labels.')
-    img_dtype: Optional[type] = Field(
+    img_dtype: Optional[str] = Field(
         None,
         description='The dtype that the training images should be cast to. '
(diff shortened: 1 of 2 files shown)


lewfish (changes requested):
Maybe I'm missing something here, but it seems like it would be simpler to just add an option to do this in the `RasterSource`. Then you wouldn't need to handle the option in multiple places and it would work for the other computer vision tasks.

AdeelH:
You mean add it to `RasterSourceConfig` and modify `RasterioSource` and `MultiRasterSource` to handle this conversion?

lewfish:
> You mean add it to `RasterSourceConfig` and modify `RasterioSource` and `MultiRasterSource` to handle this conversion?

Yes, but hopefully you would only need to modify the `RasterSourceConfig` and `RasterSource` base classes so the `get_chip` method would do the type conversion. I think that work and be simpler, but maybe there's some nuance I'm missing.

jamesmcclain:
I am late to the party, but there is now a [CastTransformer](https://github.com/azavea/noaa-flood-mapping/blob/master/pipeline/usfimr_raster_pipeline.py#L56) that you can use to cast to any type

AdeelH:
Closing this because this use case is already covered by `StatsAnalyzer` and `StatsTransformer`, so this PR is not needed.

PR #1043, opened 2020-11-14
Make Subsequent Docker Images Related

jamesmcclain:
## Overview

Cause subsequent docker images to be more closely related (saves about 5 GB per image and causes faster builds and downloads).

### DIFF
--- a/.travis/build
+++ b/.travis/build
@@ -4,5 +4,8 @@ set -e
 source .travis/env
 
 if [ "$IMAGE_TYPE" = "pytorch" ]; then
-    docker build --build-arg CUDA_VERSION="10.0" -t raster-vision-pytorch -f Dockerfile .;
+    if ! docker pull ${BASE_IMAGE}; then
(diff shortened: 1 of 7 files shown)


lewfish:
I'm not sure I understand the intention behind this PR. The steps of the Dockerfile that copy the RV source code happen at the end of the Dockerfile, so all of the time-consuming and storage intensive steps should already be cached and copied when the image is re-built.

jamesmcclain:
> I'm not sure I understand the intention behind this PR. The steps of the Dockerfile that copy the RV source code happen at the end of the Dockerfile, so all of the time-consuming and storage intensive steps should already be cached and copied when the image is re-built.

Subsequent raster-vision images on quay are mostly unrelated and differ by about 4 GB or so.

My specific use case for wanting incremental differences to be smaller (besides aesthetic and storage-related) is that I frequently develop raster-vision patches in the context of projects which I then turn into raster-vision pull requests.  As those patches are merged, I update the build for the project.

jamesmcclain:
It should also speed the CI up, too.

lewfish:
This seems useful. Right now I'm testing an alternative approach which would be simpler, but I don't know if it will work.

PR #1046, opened 2020-11-18
Add GeoDataset to allow reading directly from source imagery during training without chipping + refactor existing datasets

### LINKED ISSUE #1040: Allow reading directly from TIFFs during training
The current training workflow involves first generating a training/validation dataset from large TIFFs through the `chip` command and then doing training on the resulting dataset, treating it as if it were any other image dataset.

It should be possible to skip the chipping stage entirely and read directly from the TIFFs. This is something that the `predict` stage already does. Besides saving time and disk space, this will also allow greater flexibility during training to try experimental/ad hoc things such as:
- training with multiple window sizes
- localizing loss/metrics to geographical regions

The exact implementation details will have to be worked out, but there is already some code by @jamesmcclain to do this here: https://github.com/geotrellis/deeplab-nlcd/tree/master/src/libchips

AdeelH:
### Overview
This PR adds a new type of Dataset, `GeoDataset`, that allows the Learner to read rasters and labels directly from a `Scene` during training. `GeoDataset` is further subclassed into `SlidingWindowGeoDataset` and `RandomWindowGeoDataset`, that represent two different reading methods.

These datasets can be configured using `GeoDataConfig`, which is subclassed from `DataConfig`, and accepts a `rastervision.core.data.DatasetConfig` and either a single `GeoDataWindowConfig` Config that applies to all scenes or a `Dict[scene.id --> GeoDataWindowConfig]`.

### The Dataset classes
`pytorch_learner/dataset/*.py`

The Dataset classes are now structured as shown below. (Only classificaiton and semantic segmentation have been shown, but it's the same for object detection and regression too). _**ALL**_ datasets are now based on `AlbumentationsDataset`.
graph

### DataConfig
`pytorch_learner/learner_config.py`
`pytorch_learner/classification_learner_config.py`
`pytorch_learner/regression_learner_config.py`
`pytorch_learner/object_detection_learner_config.py`
`pytorch_learner/semantic_segmentation_learner_config.py`

To draw a distinction from `GeoData-` classes, the old classes (that work with chips) have been renamed to `ImageData-` where applicable.

### Dataset creation
`pytorch_learner/learner_config.py`
`pytorch_learner/learner.py`

Dataset creation code has been factored out into `Image-` and `Geo-` `DataConfigs` and `Learner`. And removed from the `Learner` subclasses.

### Backend config
`pytorch_backend/pytorch_*_config.py`

To avoid cluttering `PyTorckLearnerBackendConfig`, it now accepts a `DataConfig` directly instead of accepting individual fields of `DataConfig` like it did before. The examples and integration tests have been updated to make use of this.

### Other changes
- ~Adds `persist` param to `RasterioSource` that forces it to keep the file(s) open.~
- Modifies ActivateMixin to make it possible to keep a source permanently activated.
- Adds `lazy` param to `ChipClassificationLabelSource` that makes it skip label reading/inference during initialization.
- Refactors `Box`, `Scene`, `RasterSource`, `SemanticSegmentationLabelSource`, `ChipClassificationLabelSource`, `ChipClassificationLabels` to  make them work better with `GeoDataset`.

### Checklist

## Testing Instructions

* How to test this PR
  - See updated integration tests and examples.
  - ~See new examples (`isprs_potsdam_nochip.py` and `spacenet_rio_nochip.py`).~
  - See updated examples (`isprs_potsdam.py`, `spacenet_rio.py`, and `cowc_potsdam.py`).

Closes #1040


AdeelH:
Performance comparisons
image
image

jamesmcclain (commented):
This is purely ascetic, but please consider squashing together some of these commits into tighte, single-topic commits (`git rebase --interactive`).

jamesmcclain:
> This is purely aesthetic, but please consider squashing together some of these commits into tighter, single-topic commits (`git rebase --interactive`).

(Actually, it isn't purely aesthetic, it would make the PR easier to review.)

jamesmcclain (commented):
Given the user-facing API changes, it seems like these changes warrant a minor-version increment? @lewfish

jamesmcclain (changes requested):
Is `preview_batch_limit` still there?  I see it in the backwards-compatibility code (I think), but I do not see how to access it from the API.

rastervision_pytorch_backend/rastervision/pytorch_backend/examples/chip_classification/spacenet_rio.py
```diff
@@ -85,6 +86,9 @@ def make_scene(scene_info):
         train_scenes=train_scenes,
         validation_scenes=val_scenes)
 
+    img_sz = chip_sz
+    data = ClassificationImageDataConfig(img_sz=img_sz, num_workers=4)
```
AdeelH, line 90: @jamesmcclain `preview_batch_limit` goes here. e.g.

```python3
data = ClassificationImageDataConfig(img_sz=img_sz, num_workers=4, preview_batch_limit=8)
```

jamesmcclain (changes requested):
I am having trouble getting through the training stage of a previously-existing project that has been modified to use the new API (will provide details).

AdeelH:
> I am having trouble getting through the training stage of a previously-existing project that has been modified to use the new API (will provide details).

This has been fixed in the latest commit, the old example should now work without any additional changes.

jamesmcclain:
If any particles of configuration are given in two places (as now seems possible [but necessary]), I think that it makes sense to raise a validation error if the person has specified inconsistent values and raise a validation warning if they have specified consistent values in two places.  Pending that :+1:

jamesmcclain (dismissed):

lewfish (changes requested):
I agree with James that it would be easier to review if broken into a small number of single topic commits such as one for the refactorings you did, and then one for each of the changes for each computer vision task to be able to use nochips. I usually undo all my commits before making a PR and then redo them using a graphical Git tool so I can include or exclude specific lines in a commit. 

Overall this looks like a great PR.

lewfish:
> Given the user-facing API changes, it seems like these changes warrant a minor-version increment? @lewfish

I think if we were following semantic versioning it would be a major version increment, but since we're not at v1.0 yet and not following semantic versioning it would just be a minor version increment like you say. With all the changes we've made in the past few months, I think we should plan on making a release soon.

rastervision_pytorch_backend/rastervision/pytorch_backend/examples/semantic_segmentation/isprs_potsdam_nochip.py
```diff
@@ -0,0 +1,136 @@
...
+CLASS_NAMES = [
+    'Car', 'Building', 'Low Vegetation', 'Tree', 'Impervious', 'Clutter'
+]
+CLASS_COLORS = [
+    '#ffff00', '#0000ff', '#00ffff', '#00ff00', '#ffffff', '#ff0000'
+]
+
+
+def get_config(runner,
+               raw_uri: str,
```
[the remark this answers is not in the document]
AdeelH: I've actually never run the object detection example. I see that there a number of files there. Is there a guide on how to run the example?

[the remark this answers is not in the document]
lewfish: https://docs.rastervision.io/en/latest/examples.html#object-detection-cowc-potsdam-cars

AdeelH:
I've redone the commits. Hopefully they make more sense now.

AdeelH:
image

rastervision_pytorch_backend/rastervision/pytorch_backend/pytorch_learner_backend_config.py
```diff
@@ -1,94 +1,45 @@
...
-    augmentor_list, PlotOptions, Proportion)
-from rastervision.pytorch_learner.utils import validate_albumentation_transform
+    SolverConfig, ModelConfig, DataConfig, ImageDataConfig)
 
 
 @register_config('pytorch_learner_backend')
 class PyTorchLearnerBackendConfig(BackendConfig):
     model: ModelConfig
     solver: SolverConfig
+    data: DataConfig
```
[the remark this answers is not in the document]
AdeelH: Do you mean retaining all the existing fields in the learner backend config and only adding data as an optional field? That would work, but it would end up creating 2 alternate ways to specify some of the fields in the data config and we will then need some more checks to ensure that they are only specified once, which would add some additional complexity too.

I like the upgrader method because it makes the backward-compatibility-related complexity explicit and keeps it separate from the config itself.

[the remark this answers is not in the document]
AdeelH: Speaking of backward-compatibility, we only care about compatibility with previous releases and not unreleased intermediate versions, right? If so, I think we can safely move `channel_display_groups`, `img_format`, and `label_format` from the SS pipeline config to the data config. But that is probably better done in a separate PR.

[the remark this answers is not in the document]
AdeelH: It would also be nice to have a backward compatibility integration test.

lewfish:
Do you have any idea why the validation samples/sec is worse for nochip for semseg and object detection?

rastervision_pytorch_backend/rastervision/pytorch_backend/pytorch_learner_backend_config.py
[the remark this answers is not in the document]
lewfish: Ok, I think I understand things better now, and the current approach you have sounds better. Backward compatibility should be maintained for any model bundle created with any commit after (and including) the 0.12 release. This is because we may have model bundles that were created with non-release commits that need to work with the latest RV in some deployed application. I agree about the backward compatibility integration test, for which we have an issue: https://github.com/azavea/raster-vision/issues/566

AdeelH:
> Do you have any idea why the validation samples/sec is worse for nochip for semseg and object detection?

My very speculative guess would be that the single forward pass during validation is fast enough that the data reading becomes the main bottleneck, and GeoDataset is slower because of some lock-related issues when multiple workers try to read from the same file. Do you or @jamesmcclain know if GDAL uses any file level locks when reading?

rastervision_pytorch_backend/rastervision/pytorch_backend/examples/semantic_segmentation/isprs_potsdam_nochip.py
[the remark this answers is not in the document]
AdeelH: Note about object detection: I changed the `BBoxParams`'s `format` from `'coco'` to `'albumentations'`, because I was getting some strange non-deterministic errors with the former when using `A.HorizontalFlip`.

lewfish:
> My very speculative guess would be that the single forward pass during validation is fast enough that the data reading becomes the main bottleneck, and GeoDataset is slower because of some lock-related issues when multiple workers try to read from the same file. Do you or @jamesmcclain know if GDAL uses any file level locks when reading?

No, I don't know

jamesmcclain:
> > Do you have any idea why the validation samples/sec is worse for nochip for semseg and object detection?
> 
> My very speculative guess would be that the single forward pass during validation is fast enough that the data reading becomes the main bottleneck, and GeoDataset is slower because of some lock-related issues when multiple workers try to read from the same file. Do you or @jamesmcclain know if GDAL uses any file level locks when reading?

Sorry, for missing this.  Without directly answering whether GDAL uses locks while reading (different drivers do different things), I can state that reading files in GDAL is generally not thread safe (so there are some instances in which internal driver state is not properly locked).

AdeelH:
>(i) not reusing the same file handle 

I tried opening the file again with each worker, but didn't see any improvements.

AdeelH:
It seems like the performance difference between CC (Spacenet Rio) and SS/OD (Potsdam) might be due to differences in the TIF formats of their respective image files.

Spacenet Rio:
```python
{'blockxsize': 256,
 'blockysize': 256,
 'compress': 'jpeg',
 'count': 3,
 'crs': CRS({'init': 'epsg:4326'}),
 'driver': 'GTiff',
 'dtype': 'uint8',
 'height': 19584,
 'interleave': 'pixel',
 'nodata': None,
 'photometric': 'ycbcr',
 'tiled': True,
 'transform': Affine(4.487879136029412e-06, 0.0, -43.76953125,
       0.0, -4.487879136029412e-06, -22.8515625),
 'width': 19584}
```

Potsdam
```python
{'blockxsize': 608,
 'blockysize': 608,
 'compress': 'packbits',
 'count': 4,
 'crs': CRS({'init': 'epsg:32633'}),
 'driver': 'GTiff',
 'dtype': 'uint8',
 'height': 6000,
 'interleave': 'pixel',
 'nodata': None,
 'tiled': True,
 'transform': Affine(0.05, 0.0, 366976.5,
       0.0, -0.05, 5808562.6),
 'width': 6000}
```

rastervision_pytorch_backend/rastervision/pytorch_backend/examples/object_detection/cowc_potsdam.py
```diff
@@ -9,21 +9,63 @@
...
-        train_ids = train_ids[0:1]
-        val_ids = val_ids[0:1]
+        train_ids = train_ids[:2]
+        val_ids = val_ids[:2]
 
-    def make_scene(id):
+    chip_sz = 600
+    img_sz = 256
+    if nochip:
+        raster_source_persist = True
```
lewfish, line 61: It seems odd that the user needs to specify whether the raster source should be persisted. This seems like the sort of thing that should be determined automatically. I see how this could be difficult though because the raster source is constructed before the PyTorch Dataset. Maybe when the PyTorch Dataset gets the reference to the scenes, it could call a method on them which will set the persist flags.

rastervision_pytorch_backend/rastervision/pytorch_backend/examples/object_detection/cowc_potsdam.py
```diff
@@ -9,21 +9,63 @@
...
+        chip_options = ObjectDetectionChipOptions()
+    else:
+        raster_source_persist = False
+        chip_options = ObjectDetectionChipOptions(
+            neg_ratio=5.0, ioa_thresh=0.9)
```
lewfish, line 66: The `neg_ratio` is an important option for getting good performance, so there should be a way to set this for nochip as well. In general, any option that is currently set via `ChipOptions` should be settable in nochip mode. 

Come to think of it, it seems unfortunate that we have to duplicate these options in different places depending on whether you're using chips or not. Maybe we should just use the `ChipOptions` object for both cases. But perhaps it will be too difficult to do this given the current RV architecture.

lewfish (changes requested):
I think we need to make sure that the examples can be run with the same chipping options regardless of whether we're in nochip mode or not. This will be needed before the release when we'll want to run the examples in both modes and check that the accuracy hasn't gone down. 

I wish I had realized this sooner, but think that ideally there is a simpler configuration API for this whole thing. It seems like it should be possible -- from the user's standpoint -- to just say `nochip=True` in the `ChipOptions` or something and then have everything else be the same. Then, behind the scenes, it would use the `GeoDataset`, etc. I realize this will probably be too much work to do right now and might not be justifiable from a resources perspective, so we can either leave it as is, or fix it in another PR.

lewfish:
As an addendum to my last review comment: 

Put another way, the examples have a nochip argument, and based on that, different configurations are created. Ideally, we could just push the nochip option up a level into the RV source code itself, so that the logic of modifying the configuration based on nochip doesn't need to be repeated in each example. This would reduce the "degrees of freedom" in the user-facing API, and would force us to implement the same chipping options for both modes.

jamesmcclain:
> It seems like the performance difference between CC (Spacenet Rio) and SS/OD (Potsdam) might be due to differences in the TIF formats of their respective image files.
> 
> Spacenet Rio (`AOI_1_Rio/srcData/mosaic_3band/013022223132.tif`):
> 
> ```python
> {'blockxsize': 256,
>  'blockysize': 256,
>  'compress': 'jpeg',
>  'count': 3,
>  'crs': CRS({'init': 'epsg:4326'}),
>  'driver': 'GTiff',
>  'dtype': 'uint8',
>  'height': 19584,
>  'interleave': 'pixel',
>  'nodata': None,
>  'photometric': 'ycbcr',
>  'tiled': True,
>  'transform': Affine(4.487879136029412e-06, 0.0, -43.76953125,
>        0.0, -4.487879136029412e-06, -22.8515625),
>  'width': 19584}
> ```
> 
> Potsdam (`4_Ortho_RGBIR/top_potsdam_2_10_RGBIR.tif`):
> 
> ```python
> {'blockxsize': 608,
>  'blockysize': 608,
>  'compress': 'packbits',
>  'count': 4,
>  'crs': CRS({'init': 'epsg:32633'}),
>  'driver': 'GTiff',
>  'dtype': 'uint8',
>  'height': 6000,
>  'interleave': 'pixel',
>  'nodata': None,
>  'tiled': True,
>  'transform': Affine(0.05, 0.0, 366976.5,
>        0.0, -0.05, 5808562.6),
>  'width': 6000}
> ```

I would like to see the results of the experiment with duplicate validation scenes.  Your point about training being more expensive than validation and therefore perhaps hiding some of the latency is well taken, though.

rastervision_pytorch_backend/rastervision/pytorch_backend/examples/object_detection/cowc_potsdam.py
AdeelH: I agree. That makes sense.

AdeelH:
>I wish I had realized this sooner, but I think that ideally there is a simpler configuration API for this whole thing. It seems like it should be possible -- from the user's standpoint -- to just say nochip=True in the ChipOptions or something and then have everything else be the same. Then, behind the scenes, it would use the GeoDataset, etc. I realize this will probably be too much work to do right now and might not be justifiable from a resources perspective, so we can either leave it as is, or fix it in another PR.

>Come to think of it, it seems unfortunate that we have to duplicate these options in different places depending on whether you're using chips or not. Maybe we should just use the ChipOptions object for both cases. But perhaps it will be too difficult to do this given the current RV architecture.

I agree and had wanted to do something like that originally but did not want to change the RVPipeline code and deal with the added backward compatibility issues.

Ideally, in my opinion, the code should be structured such that Scenes are data sources can be consumed, using some algorithm, by the chipping stage and GeoDataset (and the analyze stage, predict stage, etc.). The algorithm should be abstracted out into something like a SceneReader that can be configured using chip/window options. But I believe we should only do this if we're willing to break backward compatibility.

rastervision_pytorch_backend/rastervision/pytorch_backend/examples/object_detection/cowc_potsdam.py
AdeelH: I have removed `persist` in favor of modifying `ActivateMixin`. The user no longer has to specify anything to change raster source behavior.

In a similar vein, `ChipClassificationLabelSourceConfig.lazy` is now set to `True` automatically. The point of the param being to disable inferring labels in the constructor (to avoid doing it twice).

AdeelH:
>I would like to see the results of the experiment with duplicate validation scenes. Your point about training being more expensive than validation and therefore perhaps hiding some of the latency is well taken, though.

I ran a segmentation experiment with 3 training and 3 validation scenes and did not notice any improvement.

rastervision_pytorch_backend/rastervision/pytorch_backend/examples/object_detection/cowc_potsdam.py
AdeelH: Done.

AdeelH: The meat of the changes is in `object_detection_dataset.py`.

rastervision_pytorch_backend/rastervision/pytorch_backend/examples/object_detection/cowc_potsdam.py
```diff
@@ -36,47 +76,84 @@ def make_scene(id):
...
+                # method=GeoDataWindowMethod.sliding,
+                method=GeoDataWindowMethod.random,
+                size=img_sz,
+                # stride=img_sz // 2,
+                size_lims=(200, 300),
+                # h_lims=(200, 300),
+                # w_lims=(200, 300),
+                max_windows=500,
+                max_sample_attempts=100,
+                ioa_thresh=0.75,
```
lewfish, line 114: I think we should make the nochip `ioa_thresh` and `neg_ratio` match the values that we were using when chipping. The performance is pretty sensitive to these options.

lewfish (changes requested):
After addressing the one comment I left I think this is ready to merge.

jamesmcclain:
:tada: :tada: :tada: :tada: :tada: :tada: :tada:

PR #1056, opened 2020-12-16
Fix bug in loading regression data

lewfish:
## Overview

This PR fixes a small bug in how regression data is loaded that surfaced after the nochip PR was merged. 

## Testing Instructions

* Tested with a pipeline in an external repo that uses the `RegressionLearner`.

### DIFF
--- a/rastervision_pytorch_learner/rastervision/pytorch_learner/dataset/dataset.py
+++ b/rastervision_pytorch_learner/rastervision/pytorch_learner/dataset/dataset.py
@@ -47,6 +47,8 @@ class AlbumentationsDataset(Dataset):
             self.normalize = False
             self.to_pytorch = False
 
+        self.transform_type = transform_type
+
     def __getitem__(self, key) -> Tuple[torch.Tensor, torch.Tensor]:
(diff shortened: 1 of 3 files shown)


rastervision_pytorch_learner/rastervision/pytorch_learner/dataset/dataset.py
```diff
@@ -47,6 +47,8 @@ def __init__(self,
             self.normalize = False
```
AdeelH, line 47: To keep it consistent with object detection, I would just add the check here and set `self.to_pytorch = False` and then modify `regression_transformer()` in `transform.py` to return `x` and `y` in the correct form.

On second thought, perhaps the `y = torch.from_numpy(y).long()` in `__getitem__` is perhaps too strong an assumption. Maybe it should just be `y = torch.from_numpy(y)` and we should expect the transform functions to supply a `y` with the correct data type.

AdeelH (changes requested):

lewfish:
I added a commit which tries to follow your second suggestion.

AdeelH (approved):

PR #1057, opened 2020-12-22
Allow saving SS predictions as probabilities

### LINKED ISSUE #1055: Allow using stride < chip size in the predict stage for semantic segmentation
This helps mitigate sharp boundaries along chip edges.

AdeelH:
## Overview

- Allows `SemanticSegmentationLabels` to hold continuous values (class probabilities).
  - Allows multiple updates for each pixel that can be averaged.
  - Allows stride < chip_sz during prediction.
- Allows `SemanticSegmentationLabelStore` to save continuous values.
  - The continuous values can optionally be quantized and saved as 1-byte uint8 (0-255) to save space.
- Allows `SemanticSegmentationLabelStore` to merge scores with scores from an existing file.
  - This allows re-running the predict step multiple times with different configurations (chip size, stride, models, etc.) to keep refining the predictions.

### Checklist

- [X] Updated `docs/changelog.rst`
- [X] Added `needs-backport` label if PR is bug fix that applies to previous minor release

## Notes
In the code, the world "smooth" is used to refer to continuous values and to distinguish them from discrete labels which are the default. I'm open to suggestion for refactoring `smooth*` to `raw*` or something else if it sounds better. I avoided `raw*` because it is used elsewhere in RV in a different context.

## Testing Instructions

* How to test this PR
  - In `isprs_potsdam.py`, make the following changes:
    ```python3
    label_store = SemanticSegmentationLabelStoreConfig(
            rgb=True,
            vector_output=[PolygonVectorOutputConfig(class_id=0)],
            smooth_output=True,
            smooth_as_uint8=True)
    ```
    ```python3
    pipeline = SemanticSegmentationConfig(
        ...,
        predict_options=SemanticSegmentationPredictOptions(stride=128),
        ...,)
    ```

Closes #1055

### DIFF
--- a/rastervision_core/rastervision/core/data/label/semantic_segmentation_labels.py
+++ b/rastervision_core/rastervision/core/data/label/semantic_segmentation_labels.py
@@ -1,4 +1,5 @@
-from typing import Tuple, Optional, List, Any, Dict
+from abc import abstractmethod
+from typing import Tuple, Optional, List, Any
 
 from rastervision.core.data.label import Labels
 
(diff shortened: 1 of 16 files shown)


lewfish:
To get rid of boundary artifacts, it looks like you are setting `stride` < `chip_size` and then averaging over multiple predictions made for a pixel. If the predictions made near the edge of a chip are bad, then shouldn't we be throwing those away rather than including them in an average?

rastervision_core/rastervision/core/data/label/semantic_segmentation_labels.py
```diff
@@ -8,75 +10,279 @@
...
+    If discrete, the labels are stored as a dict, mapping windows to label
+    arrays.
+
+    """
+
+    def __init__(self,
+                 smooth: bool = False,
+                 extent: Optional[Box] = None,
+                 num_classes: Optional[int] = None):
+        """Constructor.
```
lewfish, line 33: It seems like most of the methods follow the pattern of doing one thing if `smooth` and another if not. Maybe it would be cleaner if there were two subclasses of `SemanticSegmentationLabels` that implemented these different behaviors.

lewfish: On second thought maybe there should just be one class, but the implementation for the two modes should be unified so that for `smooth=False` it just uses a big array (with one fewer dimension) instead of the map to smaller arrays. The add method in this case would just overwrite instead of adding. Then it would make more sense to represent it as a single class, and seems like it would be simpler to understand.

lewfish (changes requested):
To get rid of boundary artifacts, it looks like you are setting stride < chip_size and then averaging over multiple predictions made for a pixel. However, if the predictions made near the edge of a chip are bad, then it seems like we should be throwing those away rather than including them in an average.

Also, I think there should be a unit test to test smooth mode for `SemanticSegmentationLabels`.

AdeelH:
>To get rid of boundary artifacts, it looks like you are setting stride < chip_size and then averaging over multiple predictions made for a pixel. However, if the predictions made near the edge of a chip are bad, then it seems like we should be throwing those away rather than including them in an average.

Boundary artifacts are not the only motivation for this. The assumption is that averaging multiple predictions for the same area (made by the model looking at that area in slightly different contexts) should improve the overall quality of the final prediction.

Besides, discarding the predictions at the edges would introduce additional complexity and another hyperparameter: the amount of boundary to discard.

rastervision_core/rastervision/core/data/label/semantic_segmentation_labels.py
AdeelH: >but the implementation for the two modes should be unified so that for smooth=False it just uses a big array (with one fewer dimension) instead of the map to smaller arrays

This is what I had originally planned to do. But if there is a situation where the windows that we will be storing labels for make up only a small portion of the whole extent, creating an array for every pixel in the extent would not be very memory efficient. I also toyed with having a separate parameter to set the format as dict/array independent of the smooth/discrete param, but that ended up making the code too convoluted.

AdeelH:
>Maybe it would be cleaner if there were two subclasses of SemanticSegmentationLabels that implemented these different behaviors.

@lewfish, I think I have succeeded in doing this. The implementation of the discrete labels is still the same as it was prior to this PR. I can change it to an array of labels if you think the implementations should be unified even if they are in different classes.

lewfish (approved):

PR #1062, opened 2021-01-06
Fix bug in semantic segmentation transform

AdeelH:
Fix for bug introduced in semantic segmentation by #1056 as a side effect of moving the type casting of `y` to before the transform is applied. Passing the mask to an albumentations transform as an `np.long` array turns it into an `np.int32`, which is incompatible with PyTorch's loss functions.

This was not caught by the semantic segmentation integration test because of `chip_sz` being equal to `img_sz` which causes the `Resize` transform to do nothing and return the mask as is. To reproduce the error, set `img_sz` to something different (in `integration_tests/semantic_segmentation/config.py`).

### DIFF
--- a/rastervision_pytorch_learner/rastervision/pytorch_learner/dataset/transform.py
+++ b/rastervision_pytorch_learner/rastervision/pytorch_learner/dataset/transform.py
@@ -23,10 +23,11 @@ def classification_transformer(inp: Tuple[Any, Any],
                                ) -> Tuple[np.ndarray, np.ndarray]:
     """Apply transform to image only."""
     x, y = inp
-    x, y = np.array(x), np.array(y, dtype=np.long)
+    x, y = np.array(x), np.array(y)
     if transform is not None:
         out = transform(image=x)
         x = out['image']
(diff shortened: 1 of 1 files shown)


rastervision_pytorch_learner/rastervision/pytorch_learner/dataset/transform.py
```diff
@@ -23,10 +23,11 @@ def classification_transformer(inp: Tuple[Any, Any],
                                ) -> Tuple[np.ndarray, np.ndarray]:
     """Apply transform to image only."""
     x, y = inp
-    x, y = np.array(x), np.array(y, dtype=np.long)
+    x, y = np.array(x), np.array(y)
     if transform is not None:
         out = transform(image=x)
         x = out['image']
+    y = y.astype(np.long)
```
AdeelH, line 30: This is not necessary for classification (because `y` is not modified by the transform), but moving the casting to the end emphasizes that the returned `y` should be a `long`.

lewfish (approved):

PR #1065, opened 2021-01-11
Set runtime flag according to docker version

ashnair1:
Running `docker/run  --gpu` fails with docker>=19.03 the following error:

```
docker: Error response from daemon: OCI runtime create failed: unable to retrieve OCI runtime error (open /run/containerd/io.containerd.runtime.v1.linux/moby/723d7848a6cf4e0c20006d6f3e0cc4b3ccc01be12b468e61235e8b54bc6771c0/log.json: no such file or directory): exec: "nvidia-container-runtime": executable file not found in $PATH: : unknown.
```
This is due to docker deprecating the `runtime` argument post v19.03. This PR addresses this by first checking the docker version and setting runtime flag accordingly


lewfish (approved):

lewfish:
Thank you!

PR #1086, opened 2021-02-01
Refactor prediction pipeline. Reroute prediction input through a Dataset/DataLoader so that the base_transform can be applied to it.

### LINKED ISSUE #1053: Apply base_transform in the predict stage
Currently, the predict stage uses an entirely different code path than validation, and therefore, the input does not go through the `base_transform` transform (which might include things like resizing and normalization). This means that the input is not guaranteed to be in the form that the model expects.

AdeelH:
## Overview

This PR
- moves the bulk of the prediction logic from `RVPipeline` to the learner backend
  - where a `SlidingWindowGeoDataset` is used to traverse the scene
  - this dataset is created via the `Learner`, so has access to the `base_transform` transform
- makes prediction params (`chip_sz`, `batch_sz`, `stride`) configurable using `RVPipeline.predict_options`
  - these override `predict_chip_sz` and `predict_batch_sz` if specified
- adds an `iter_predictions()` method to the `Learner` that returns a generator of predictions from a `DataLoader`
  - motivation: avoid possibly storing all the predictions of a scene in memory twice (once as the return values of `predict_dataloader()` and then as `Labels`)
- adds a `__setitem__` to the `Labels` subclasses for ease of use with the new prediction code
- adds `--ipc=host` to the `docker run` command used for CI testing (`.travis/test`)

### Checklist

Closes #1053

### DIFF
--- a/.travis/test
+++ b/.travis/test
@@ -6,5 +6,5 @@ if [ "$IMAGE_TYPE" = "pytorch" ]; then
     docker run -w $(pwd) -v $(pwd):$(pwd) --rm -it ${PYTORCH_IMAGE} rm -f $(pwd)/.coverage $(pwd)/coverage.xml
     docker run --rm -it ${PYTORCH_IMAGE} /opt/src/scripts/style_tests
     docker run --rm -it ${PYTORCH_IMAGE} /opt/src/scripts/unit_tests
-    docker run --rm -it ${PYTORCH_IMAGE} /opt/src/scripts/integration_tests
+    docker run --rm -it --ipc=host ${PYTORCH_IMAGE} /opt/src/scripts/integration_tests
 fi
--- a/rastervision_core/rastervision/core/data/label/chip_classification_labels.py
+++ b/rastervision_core/rastervision/core/data/label/chip_classification_labels.py
@@ -1,3 +1,7 @@
+from typing import Tuple, Union
+
+import numpy as np
+
 from rastervision.core.box import Box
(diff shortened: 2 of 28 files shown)


lewfish (commented):
We talked about this PR in person, but I want to record some notes about this for later:
* It seems like this PR was bigger than it needed to be to address the original issue which was applying the same transformation at prediction time as during training. 
* Refactoring to remove the prediction logic from the `core` package and doing it in the `pytorch_backend` and `pytorch_learner` using the `GeoDataset` has the disadvantage that this logic needs to be replicated in any new backend that is added. On the other hand, it seems advantageous that we can re-use some of the sliding window logic that's in the `GeoDataset`. By making predictions using the `DataLoader` which has support for multiple threads, we could potentially get a speed improvement.
* If we were to keep the current PR, we would need to modify `Backend` so that the signature of the `predict` method is in sync with the one in `PytorchLearnerBackend`

AdeelH:
See #1112 for a minimal fix for #1053.

> It seems like this PR was bigger than it needed to be to address the original issue which was applying the same transformation at prediction time as during training.

Agreed. This PR is additionally addressing an additional concern not currently captured in an issue. Namely that the 'pipelines' through which data is fed into the model during validation and prediction are different and must be maintained separately to ensure that the processing they do on the data is equivalent. This means that things like the application of transforms, normalization, conversion to PyTorch etc. (and any changes to these in the future) have to be duplicated.

AdeelH:
Stale.

PR #1144, opened 2021-03-24
Fix image plot by adding default plot transform

### LINKED ISSUE #1132: Input images don't display in dataloader and test_preds images
When running the `tiny_spacenet.py` example in 0.13 with `rastervision run local rastervision.pytorch_backend.examples.tiny_spacenet`, the input images are blank in `train/dataloaders/*.png` and `train/test_preds.png`. 

train

test_preds

lewfish:
## Overview

This PR fixes blank image plots by adding a default plot transform which normalizes the pixel values so that the minimum and maximum values are 0 and 1, respectively. This is to handle cases where the pixel values are small and cannot be seen, such as when a uint16 image with values close to 0 is normalized to 0-1. An example of this occurring is in the SpaceNet examples. 

### Notes

* Originally, I thought I would detect that the values were outside the 0-1 range and then correct them if needed. However, the the values are already within the 0-1 range and are just tiny.
* It's unclear how to detect that values are too small to see, so it seemed most straightforward to just scale things to be between 0 and 1. That will distort the image in some cases, but at least it should always be visible. If people really want, they can always set the `plot_transform` to None. 
* After this is merged we should make a bug fix release.

## Testing Instructions

* I inspected the plots for the following examples which seems to run the gamut: SpaceNet Vegas (uint16) with nochip=True and False, and COWC Potsdam (uint8) with nochip=True and False. The image plots are visible. Here are plots for the SpaceNet Vegas example which was the original motivation for this PR:
train
* I added a unit test for the `MinMaxTransform` which I added.

Closes #1132

### DIFF
--- a/rastervision_pytorch_learner/rastervision/pytorch_learner/semantic_segmentation_learner.py
+++ b/rastervision_pytorch_learner/rastervision/pytorch_learner/semantic_segmentation_learner.py
@@ -215,12 +215,6 @@ class SemanticSegmentationLearner(Learner):
             x = tf(image=x.numpy())['image']
             x = torch.from_numpy(x)
 
-        # apply transform, if given
-        if self.cfg.data.plot_options.transform is not None:
-            tf = A.from_dict(self.cfg.data.plot_options.transform)
-            x = tf(image=x.numpy())['image']
-            x = torch.from_numpy(x)
-
         for i in range(batch_sz):
             ax = (fig, axes[i])
             if z is None:
--- a/rastervision_pytorch_learner/rastervision/pytorch_learner/semantic_segmentation_learner.py
+++ b/rastervision_pytorch_learner/rastervision/pytorch_learner/semantic_segmentation_learner.py
@@ -212,8 +212,8 @@ class SemanticSegmentationLearner(Learner):
         # apply transform, if given
         if self.cfg.data.plot_options.transform is not None:
             tf = A.from_dict(self.cfg.data.plot_options.transform)
-            x = tf(image=x.numpy())['image']
-            x = torch.from_numpy(x)
+            imgs = [tf(image=img)['image'] for img in x.numpy()]
+            x = torch.from_numpy(np.stack(imgs))
 
         for i in range(batch_sz):
             ax = (fig, axes[i])
--- a/rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
+++ b/rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
@@ -950,8 +950,8 @@ class Learner(ABC):
         # apply transform, if given
         if self.cfg.data.plot_options.transform is not None:
             tf = A.from_dict(self.cfg.data.plot_options.transform)
-            x = tf(image=x.numpy())['image']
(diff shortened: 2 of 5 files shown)


rastervision_pytorch_learner/rastervision/pytorch_learner/semantic_segmentation_learner.py
```diff
@@ -209,12 +209,6 @@ def plot_batch(self,
...
 
-        # apply transform, if given
-        if self.cfg.data.plot_options.transform is not None:
-            tf = A.from_dict(self.cfg.data.plot_options.transform)
-            x = tf(image=x.numpy())['image']
-            x = torch.from_numpy(x)
-
         # apply transform, if given
         if self.cfg.data.plot_options.transform is not None:
             tf = A.from_dict(self.cfg.data.plot_options.transform)
```
AdeelH, line 214: I think ideally the transform should be applied on individual images in `x`, because albumentation transforms usually only work on single images rather than batches. This doesn't seem to matter in the case of normalization, but might cause problems if the user specifies some other transformation.

Something like:
```python
        if self.cfg.data.plot_options.transform is not None:
            tf = A.from_dict(self.cfg.data.plot_options.transform)
            x = x.numpy()
            imgs = [tf(image=img)['image'] for img in x]
            x = torch.from_numpy(np.stack(imgs))
```

AdeelH:
Do the images look a bit too green to you too? I was getting similar output when I took a small stab at this some days ago and thought I might be missing something.

Update: I just tried doing the normalization per channel, and the results look more realistic, at least for `tiny_spacenet`.

**Before**
image

**After**
image

Code:
```python
    def _apply_on_channel(self, image, **params):
        out = cv2.normalize(
            image,
            None,
            self.min_val,
            self.max_val,
            cv2.NORM_MINMAX,
            dtype=self.dtype)
        # We need to clip because sometimes values are slightly less or more than
        # min_val and max_val due to rounding errors.
        return np.clip(out, self.min_val, self.max_val)

    def apply(self, image, **params):
        if image.ndim <= 2:
            return self._apply_on_channel(image, **params)

        assert image.ndim == 3

        chs = [
            self._apply_on_channel(ch, **params)
            for ch in image.transpose(2, 0, 1)
        ]
        out = np.stack(chs, axis=2)
        return out
```

rastervision_pytorch_learner/rastervision/pytorch_learner/semantic_segmentation_learner.py
lewfish: Yup, you're right.

lewfish:
> ```python
>         if image.ndim <= 2:
>             return self._apply_on_channel(image, **params)
> ```

Why is this <= 2? Shouldn't it be < 2 (ie. a single channel image)?

AdeelH:
An image of shape (H, W) would have `ndim` = 2

lewfish:
> Do the images look a bit too green to you too? I was getting similar output when I took a small stab at this some days ago and thought I might be missing something.

Good call, that looks better with per channel normalization. I pushed an update.

AdeelH (approved):

PR #1146, opened 2021-03-26
Increment RV version to 0.13.1

lewfish:


AdeelH:
@lewfish shouldn't this have been merged into master as well? There's some weirdness going on with dependency resolution in doc builds (e.g. https://readthedocs.org/projects/raster-vision/builds/17229095/) seemingly due to the mismatch between the version number in master (0.13) and the one on pypi (0.13.1).

Does it make sense to change the version numbers in all `setup.py`s to 0.13.1?

lewfish:
> @lewfish shouldn't this have been merged into master as well? There's some weirdness going on with dependency resolution in doc builds (e.g. https://readthedocs.org/projects/raster-vision/builds/17229095/) seemingly due to the mismatch between the version number in master (0.13) and the one on pypi (0.13.1).
> 
> Does it make sense to change the version numbers in all `setup.py`s to 0.13.1 on master?

The docs for the release process say not to push this to `master`. (See step 3 in https://docs.rastervision.io/en/0.13/release.html#bug-fix-release.) But, I don't see the reason for this. I think it makes more sense to push it to `master` especially if it fixes the build.

PR #1151, opened 2021-04-21
Add within argument to filter_by_aoi

lewfish:
This adds a `within` argument to the `Box.filter_by_aoi` method, which is used in another project to include windows that merely intersect with the AOIs. It doesn't really make sense to add this an option to the config for classification, semantic segmentation, or object detection.

### DIFF
--- a/rastervision_core/rastervision/core/box.py
+++ b/rastervision_core/rastervision/core/box.py
@@ -4,7 +4,7 @@ import math
 import random
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


AdeelH:
How about an `ioa_threshold` param instead of `within` for deciding which windows to keep? It would allow avoiding windows with very small intersections with the AOI.

AdeelH (approved):

PR #1152, opened 2021-04-22
Add some options needed for SpaceNet SSL experiments

lewfish:
This PR adds some options that were needed for running experiments on SpaceNet using SSL. These were used in https://github.com/lewfish/ssl/blob/main/spacenet/spacenet_ssl.py

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -114,6 +114,3 @@ deployment/ansible/*.retry
 ./data
 ./notebooks
 .vscode/
-
-# dir that is mounted to /opt/data
-data/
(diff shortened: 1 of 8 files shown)


.gitignore
```diff
@@ -114,6 +114,3 @@ deployment/ansible/*.retry
 ./data
 ./notebooks
```
AdeelH, line 115: These actually do not seem to work. My understanding is that `/data/` or `data/` will ignore the `data` directory, while `/data` will additionally ignore any file named `data`.

I think `/data/` and `/notebooks/` is what we want.

AdeelH (approved):
I didn't realize I hadn't approved this.

PR #1218, opened 2021-07-21
Migrate from Travis to GitHub Actions

### LINKED ISSUE #1129: Migrate from Travis to GitHub Actions
We've been moving away from using Travis for CI across the organization. We've been investing heavily in GitHub Actions instead. We should migrate this repository to use Actions for CI and building Docker images.

Required by #1130.

colekettler:
## Overview

Ports Travis CI config to a GitHub Actions workflow.

### Checklist

## Testing Instructions

 * Verify that the [GitHub Actions CI workflow](https://github.com/azavea/raster-vision/runs/3146625884) passes.
* Verify that [this Release workflow](https://github.com/azavea/raster-vision/runs/3146630550) ran.
* Verify that the Release workflow has uploaded a working Docker image to Quay.io:
  ```#bash
  docker pull quay.io/azavea/raster-vision:pytorch-test-gha
  ```
* Run `./scripts/test` locally and confirm that enabled tests run successfully and all pass.
  * Two semantic segmentation integration tests have been failing on `master` and have been disabled for now.
* Verify that a code coverage report has been successfully uploaded to [Codecov](https://codecov.io/gh/azavea/raster-vision/commit/cdf0ab2d573a447903e52bb9b1ce41e6a309c557/graphs).
* Confirm that Release workflow logic covers all versions previously pushed by `.travis/deploy`.

Closes #1129


colekettler:
☝🏻 Fancy!

.coveragerc
```diff
@@ -1,10 +1,7 @@
...
 omit =
-    rastervision/protos/**.py
-    rastervision/utils/filter_geojson.py
-    rastervision/utils/geojson.py
-    rastervision/utils/misc.py
-    rastervision/data/vector_source/label_maker/**.py
-    rastervision/data/label/tfod_utils/**.py
+    */rastervision_core/rastervision/core/data/vector_source/label_maker/*.py
+    */rastervision_core/rastervision/core/data/label/tfod_utils/*.py
+    */rastervision_core/rastervision/core/utils/*.py
```
colekettler, line 7: Some of the originally excluded file patterns no longer exist. I've tried to keep the current exclusions in the spirit of the original, ignoring vendored code and non-core utilities. I expect folks more familiar with which code paths are important can tweak this over time if we decide to keep up code coverage reporting.

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,21 @@
...
+on:
+  pull_request:
+
+jobs:
+  build:
+    name: build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        image_type: [pytorch]
```
colekettler, line 12: I recognize that specifying a separate variant is likely a holdover from the transition from v1 to v2. However it's a pretty easy thing to interpolate in from the build matrix and it buys us a little extra flexibility later on. If this isn't actually useful I can hardcode it into the build and publish scripts instead.

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,21 @@
...
+    env:
+      DOCKER_BUILDKIT: 1
+      IMAGE_TYPE: ${{ matrix.image_type }}
+    steps:
+      - uses: actions/checkout@v2
+
+      - run: ./scripts/cibuild
+
+      - uses: codecov/codecov-action@v2
```
colekettler, line 21: Turns out Codecov [publishes an action](https://github.com/codecov/codecov-action), so we don't have to script this ourselves anymore. It works well!

We also don't need to maintain an upload token since this is a public repo.

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - master
+      - "[0-9]+" # Major version only
+      - "[0-9]+.[0-9]+" # Major and minor version
+      - "[0-9]+.[0-9]+.[0-9]+" # Major, minor, and patch version
```
colekettler, line 9: This should cover release branches like `0.12`, `0.12.1`, etc.

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+    name: release
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        image_type: [pytorch]
+    env:
+      DOCKER_BUILDKIT: 1
+      IMAGE_TYPE: ${{ matrix.image_type }}
+      QUAY_USER: ${{ secrets.QUAY_RASTERVISION_USER }}
+      QUAY_PASSWORD: ${{ secrets.QUAY_RASTERVISION_PASSWORD }}
```
colekettler, line 24: I've added this secret to our organization and shared it with this repo.

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - run: ./scripts/cibuild
+
+      - uses: codecov/codecov-action@v2
```
colekettler, line 30: There is some duplication here with the `CI` workflow. It ended up being really tricky to filter and disambiguate semver release branches and PRs without getting into some more complicated property checks and regex magic. Splitting this out into a separate workflow felt clearer.

The way things are structured, PRs will only be built and have a test suite run on them with the `CI` workflow, and pushes to master / release branches / tags will have the same thing done plus pushing images to Quay.io. Everything only needs to get built and tested once per event. The separate workflows won't slow down overall build time or waste build minutes.

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - run: |
+          echo "GIT_BRANCH=${GITHUB_REF##*/}" >> $GITHUB_ENV
+          echo "GIT_COMMIT=${GITHUB_SHA:0:7}" >> $GITHUB_ENV
```
colekettler, line 34: This is the approach [recommended by GitHub Actions](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable) to add environment variables. The `GIT_BRANCH` variable will snag the branch name or tag, and the `GIT_COMMIT` variable gets the short commit hash for image tagging.

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - run: ./scripts/cipublish
+        if: github.ref == 'refs/heads/master'
+        env:
+          IMAGE_VERSION: ${{ env.GIT_COMMIT }}
+
+      - run: ./scripts/cipublish
+        if: github.ref == 'refs/heads/master'
+        env:
+          IMAGE_VERSION: latest
```
colekettler, line 49: If we're building a commit on `master`, we push an image tagged with both the short commit hash and `latest`.

.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+      - uses: codecov/codecov-action@v2
+
+      - run: |
+          echo "GIT_BRANCH=${GITHUB_REF##*/}" >> $GITHUB_ENV
+          echo "GIT_COMMIT=${GITHUB_SHA:0:7}" >> $GITHUB_ENV
+
+      - run: ./scripts/cipublish
+        if: github.ref != 'refs/heads/master'
+        env:
+          IMAGE_VERSION: ${{ env.GIT_BRANCH }}
```
colekettler, line 39: If we're building a release branch or tag, we push an image tagged with that branch or tag version.

.readthedocs.yml
```diff
@@ -19,7 +19,7 @@ formats:
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
-  version: 3.6
+  version: 3.7
```
colekettler, line 22: There were a few instances where we were still using 3.6, so I updated it to 3.7 across the board.

Dockerfile
```diff
@@ -26,7 +26,7 @@ RUN python -m pip install --upgrade pip
 RUN conda install -y -c conda-forge gdal=3.0.4
 
 # Setup GDAL_DATA directory, rasterio needs it.
-ENV GDAL_DATA=/opt/conda/lib/python3.6/site-packages/rasterio/gdal_data/
+ENV GDAL_DATA=/opt/conda/lib/python3.7/site-packages/rasterio/gdal_data/
```
colekettler, line 29: There were a few instances where we were still using 3.6, so I updated it to 3.7 across the board.

docs/release.rst
```diff
@@ -98,9 +98,9 @@ Bug Fix Release
...
-#. Make and merge a PR against ``0.8`` (but not ``master``) that increments the version in each ``setup.py`` file to ``0.8.1``. Then wait for the ``0.8`` branch to be built by Travis and the ``0.8`` Docker images to be published to Quay. If that is successful, we can proceed to the next steps of actually publishing a release.
+#. Make and merge a PR against ``0.8`` (but not ``master``) that increments the version in each ``setup.py`` file to ``0.8.1``. Then wait for the ``0.8`` branch to be built by GitHub Actions and the ``0.8`` Docker images to be published to Quay. If that is successful, we can proceed to the next steps of actually publishing a release.
 #. Using the Github UI, make a new release. Use ``0.8.1`` as the tag, and the ``0.8`` branch as the target.
-#. The image for ``0.8`` is created automatically by Travis, but we need to manually create images for ``0.8.1``. For this you will need an account on Quay  under the Azavea organization.
+#. The image for ``0.8`` is created automatically by GitHub Actions, but we need to manually create images for ``0.8.1``. For this you will need an account on Quay  under the Azavea organization.
```
colekettler, line 103: Is this still true? It looks like we should automatically build and push minor version images just fine.

integration_tests/integration_tests.py
```diff
@@ -45,20 +45,21 @@
...
+    #         'nochip': False
+    #     }
+    # },
+    # 'semantic_segmentation.nochip': {
+    #     'task': 'semantic_segmentation',
+    #     'module': 'integration_tests.semantic_segmentation',
+    #     'kwargs': {
+    #         'nochip': True
+    #     }
+    # }
```
colekettler, line 62: These tests are failing on `master` as well. I've disabled them for now and created an issue to track fixing them later.

rastervision_pytorch_backend/rastervision/pytorch_backend/examples/chip_classification/spacenet_rio.py
```diff
@@ -174,7 +174,11 @@ def make_scene(scene_info) -> SceneConfig:
         external_loss_def=external_loss_def)
 
     backend = PyTorchChipClassificationConfig(
-        data=data, model=model, solver=solver, test_mode=test, log_tensorboard=True,
+        data=data,
+        model=model,
+        solver=solver,
+        test_mode=test,
+        log_tensorboard=True,
```
colekettler, line 181: Formatting fix to satisfy the format checker.

requirements-dev.txt
```diff
@@ -1,7 +1,7 @@
 flake8==3.5.*
 moto==2.0.10
-coverage==4.5.1
-codecov==2.0.15
+coverage==5.5
+codecov==2.1.11
```
colekettler, line 4: Coverage fixed a few bugs since we last updated it, so an upgrade was necessary to get things working with Python 3.7 again.

scripts/cibuild
```diff
@@ -0,0 +1,28 @@
...
+
+if [[ ${BASH_SOURCE[0]} == "$0" ]]; then
+    if [[ ${1:-} == "--help" ]]; then
+        usage
+    else
+        docker build \
+            --build-arg CUDA_VERSION="10.2" \
+            -t "raster-vision-${IMAGE_TYPE}" \
+            -f Dockerfile \
+            .
```
colekettler, line 24: I've generally removed conditional checks for whether we're building a pytorch image and interpolated in the `IMAGE_TYPE` instead. If we introduce other variant types later on and need to parameterize other things like the Cuda version, we can add it to the build matrix.

scripts/test
```diff
@@ -0,0 +1,57 @@
...
+}
+
+if [[ ${BASH_SOURCE[0]} == "${0}" ]]; then
+    if [[ ${1:-} == "--help" ]]; then
+        usage
+    else
+        if [[ -z ${CI} ]]; then
+            # Local test suite runs against pytorch image by default
+            IMAGE_TYPE=${IMAGE_TYPE:-pytorch}
+        fi
```
colekettler, line 23: This script is run in CI to run the full test suite. It can also be used locally for sake of convenience, and will default to using the pytorch variant.

colekettler:
@lewfish Tagging you for review since you're familiar with the existing Docker image release workflow. I wanted to make sure I covered all the bases and that I'm not disrupting your local workflow!

Happy to discuss if you have questions about any of the GitHub Actions stuff. It replaces most Travis features pretty seamlessly but it is structured differently. I'll also look over the config with Rocky once he's available.

lewfish (approved):
Nice job!

docs/release.rst
colekettler: Cool, I'll remove that bit from the documentation. Thanks!

scripts/test
```diff
@@ -0,0 +1,52 @@
...
+}
+
+if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
+    if [ "${1:-}" = "--help" ]; then
+        usage
+    else
+        PYTORCH_IMAGE=raster-vision-pytorch
+
+        # Delete old coverage reports
+        docker run \
```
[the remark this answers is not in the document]
colekettler: Good callout, that's an important distinction.

Our typical convention is to merge them all into `scripts` and convert them to [STRTA semantics](https://github.com/azavea/architecture/blob/master/doc/arch/adr-0000-scripts-to-rule-them-all.md). It's pretty common for scripts to run either on the host or inside of a Docker container depending on whether they're being run locally vs. on CI. `docker/build` would probably become `scripts/update` and `docker/run` would probably become `scripts/server`.

I'm hesitant to recommend that change without some further discussion though. The consistency that STRTA brings is incredibly useful, but its semantics were designed in the context of a web application. There's no server here so running `scripts/server` is kind of misleading, especially for an open source project that may have contributors outside of Azavea. It's also worth mentioning that STRTA was designed with Vagrant VMs in mind which were assumed to provide Docker and any necessary build dependencies without requiring any manual installation. That also isn't the case here.

This seems to be part of a larger thread of some of our web app conventions not quite fitting the bill for library and processing pipeline projects. I'll queue this up for discussion with Rocky.

In the short term I could add a check for whether the `test` script is running in a CI environment. If so, run tests in containers. If not, it's just an alias for the full test suite run outside of Docker on the host. We do this commonly and it would preserve existing conventions when run outside of CI. Does that sound good?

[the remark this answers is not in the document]
lewfish: > In the short term I could add a check for whether the `test` script is running in a CI environment. If so, run tests in containers. If not, it's just an alias for the full test suite run outside of Docker on the host. We do this commonly and it would preserve existing conventions when run outside of CI. Does that sound good?

That sounds good. 

> I'm hesitant to recommend that change without some further discussion though. The consistency that STRTA brings is incredibly useful, but its semantics were designed in the context of a web application. There's no server here so running `scripts/server` is kind of misleading, especially for an open source project that may have contributors outside of Azavea. It's also worth mentioning that STRTA was designed with Vagrant VMs in mind which were assumed to provide Docker and any necessary build dependencies without requiring any manual installation. That also isn't the case here.

On second thought I would strongly prefer to not rename or move any of the existing scripts. We have a lot of other repos that use this one as a template and there's also a lot of muscle memory on the script names.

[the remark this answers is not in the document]
colekettler: Cool, I'll get the `test` script updated and won't touch the Docker scripts. I've brought up STRTA semantics as they pertain to libraries with Rocky. That'll be an ongoing discussion between ops and library dev folks. We'll work out a common convention that makes sense for everybody.

docs/release.rst
colekettler: After reading this more closely, it is still technically true. With how we currently tag releases, we will automatically release images for `0.8` based on the branch, and `v0.8.1` based on our tags prefixed with `v`, but since there's no `0.8.1` branch or tag, that will still need to be done manually.

Is having both `0.8.1` and `v0.8.1` actually valuable though? I do see both tags present on Quay.io, so we've been doing it, but I don't know if there was intent behind that.

lewfish: I don't think that having both 0.8.1 and v0.8.1 is valuable. If we are automatically releasing the bug fix versions based on tags that's good enough.

lewfish:
I fixed https://github.com/azavea/raster-vision/issues/1220 in https://github.com/azavea/raster-vision/pull/1236

.github/workflows/release.yml
rbreslow: Do we know what pattern matching engine is used here? May be more semantically correct to try https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string.

.github/workflows/release.yml
rbreslow: I'd consider making these secrets accessible only to the `cipublish` step. For example:

https://github.com/rbreslow/docker-steamcmd/blob/746d7db506a833af9a6490a16bc09b4e5a86b6ba/.github/workflows/continuous_integration.yml#L21-L25

.github/workflows/release.yml
rbreslow: This seems like a reasoned tradeoff.

.github/workflows/release.yml
rbreslow: Consider allowing `GIT_COMMIT` to be sourced by the usual:

```bash
if [[ -n "${GIT_COMMIT}" ]]; then
    GIT_COMMIT="${GIT_COMMIT:0:7}"
else
    GIT_COMMIT="$(git rev-parse --short HEAD)"
fi
```

This works on GitHub Actions, since the Ubuntu runner has `git` installed. Also, consider determining the branch name in a similar fashion.

scripts/cibuild
rbreslow, line 24: Did you consider describing the image in a `docker-compose.yml` file?

scripts/test
```diff
@@ -0,0 +1,60 @@
...
+                "raster-vision-${IMAGE_TYPE}" \
+                /opt/src/scripts/integration_tests
+
+            # Create new coverage reports
+            docker run \
+                -w "$(pwd)" \
+                -v "$(pwd):$(pwd)" \
+                --rm -t \
+                "raster-vision-${IMAGE_TYPE}" \
+                coverage xml
```
rbreslow, line 43: Ditto. I think this stuff would be a lot more clear if we defined the working directories, volume mounts, TTY allocation, etc. in a `docker-compose.yml` file.

scripts/cipublish
```diff
@@ -0,0 +1,20 @@
...
+        "Usage: $(basename "$0")
+Publish container images to Quay.io.
+"
+}
+
+docker login -u="${QUAY_USER}" -p="${QUAY_PASSWORD}" quay.io
+
+docker tag "raster-vision-${IMAGE_TYPE}" \
+    "quay.io/azavea/raster-vision:${IMAGE_TYPE}-${IMAGE_VERSION}"
+docker push "quay.io/azavea/raster-vision:${IMAGE_TYPE}-${IMAGE_VERSION}"
```
rbreslow, line 20: Why the lack of our usual check for script execution vs. sourcing (e.g., `${BASH_SOURCE[0]} == "${0}"`)?

rbreslow (commented):
Great work. Just a few comments on semantics to start.

scripts/test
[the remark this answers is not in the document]
rbreslow: https://github.com/azavea/architecture/issues/10

integration_tests/integration_tests.py
colekettler: I saw these got fixed in #1236 so I dropped the commit and rebased onto `master`.

.github/workflows/release.yml
colekettler: I don't recall what the specific engine is, but I know it's not a PCRE engine and doesn't support capture groups or anything fancy like that: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

Wildcard characters, digit ranges, and n-or-more matchers are about as far as it goes.

.github/workflows/release.yml
colekettler: The catch here is that I'd like for it to be available to the workflow itself, and then have it passed along to `cipublish` as the `IMAGE_VERSION` variable. That lets us condition what kind of tag we want to push in the workflow instead of needing to pass the ref into `cipublish` and then do branching logic in there to determine whether the tag should be the branch name, short commit hash, or `latest`.

We'd also need to decide if we want to push two tags in `cipublish` if we're on `master` (hash and `latest`), either in the script or by duplicating a check in the workflow config. That's kinda what I was trying to get away from in the old `.travis/deploy` script. Calling a simpler script twice with different arguments feels less complicated than a script that needs to have some nested conditional branches.

We could  definitely do it that way, but I feel that keeping the branching logic closer to the Actions context it conditions on and having `cipublish` only ever do one thing, publish with the provided tag, is easier to read and understand. What are your thoughts?

scripts/cipublish
colekettler: Just an oversight here, I'll get that added in. Thanks for the catch!

scripts/test
colekettler: I agree! I've thought about this a little more and I'd like to do this in a separate PR. We'd need to touch some other scripts like `docker/run` and `docker/build` if we want to avoid duplicating config, and we'd also need to introduce another development dependency (at least until Compose v2 is out of beta and is available with any recent version of Docker Engine). That's going to increase the diff a good bit and will require another more thorough round of local testing to make sure I'm not breaking anybody's workflow. That's something I'd like to make very visible as a separate decision from this migration.

I can get a follow-up issue set up prior to merging. What do you think?

integration_tests/integration_tests.py
colekettler: @lewfish Heads up, I did run into a test failure here for `semantic_segmentation.basic`. Log lines in question are [here](https://github.com/azavea/raster-vision/runs/3284883787#step:3:5042).

Weirdly, it passed just fine on a subsequent run. Just wanted to raise that to your attention in case you run into flakey results later on.

colekettler:
@rbreslow I responded to your comments, coming back to you!

scripts/test
rbreslow: Yeah, I'm fine with a follow up issue that outlines concerns.

.github/workflows/release.yml
rbreslow: > The catch here is that I'd like for it to be available to the workflow itself, and then have it passed along to `cipublish` as the `IMAGE_VERSION` variable. That lets us condition what kind of tag we want to push in the workflow instead of needing to pass the ref into `cipublish` and then do branching logic in there to determine whether the tag should be the branch name, short commit hash, or `latest`.

This makes sense 👍 .

> [...] That's kinda what I was trying to get away from in the old `.travis/deploy` script. Calling a simpler script twice with different arguments feels less complicated than a script that needs to have some nested conditional branches.

On the contrary, we've used conditional logic in `cipublish` [for things like](https://github.com/geotrellis/geotrellis-server/blob/01b01224545a36ad58731c6d1c19843212137656/scripts/cipublish#L20-L24) Scala artifact publishing.

> We could definitely do it that way, but I feel that keeping the branching logic closer to the Actions context it conditions on and having `cipublish` only ever do one thing, publish with the provided tag, is easier to read and understand.

While I agree that having `cipublish` do only one thing is easier to read, I disagree that it is easier to understand. `cipublish` is supposed to wrap the workflow so that I could run `cipublish` on my laptop, and everything would work. Right now, it wouldn't work, and understanding the workflow requires you to look in two places.

What are your thoughts on this?

```bash
if [[ -n "${GITHUB_SHA}" ]]; then
    GITHUB_SHA="${GITHUB_SHA:0:7}"
else
    GITHUB_SHA="$(git rev-parse --short HEAD)"
fi

if [[ -n "${GITHUB_REF}" ]]; then
    GITHUB_REF="${GITHUB_REF}"
else
    # We don't know where GitHub sources the value of GITHUB_REF. git-describe
    # is close enough to be valid for publishing.
    GITHUB_REF="$(git describe --all)"
fi
```

```bash
docker login -u ${QUAY_USER} -p ${QUAY_PASSWORD} quay.io

if [[ "${GITHUB_REF}" == "refs/heads/master" ]]; then
    # On master, use the current SHA and latest
    docker tag "raster-vision-${IMAGE_TYPE}" \
        "quay.io/azavea/raster-vision:${IMAGE_TYPE}-${GITHUB_SHA}"
    docker push "quay.io/azavea/raster-vision:${IMAGE_TYPE}-${GITHUB_SHA}"

    docker tag "raster-vision-${IMAGE_TYPE}" \
        "quay.io/azavea/raster-vision:${IMAGE_TYPE}-latest"
    docker push "quay.io/azavea/raster-vision:${IMAGE_TYPE}-latest"
else
    # For everything else, use the branch or tag name
    docker tag "raster-vision-${IMAGE_TYPE}" \
        "quay.io/azavea/raster-vision:${IMAGE_TYPE}-${GITHUB_REF##*/}"
    docker push "quay.io/azavea/raster-vision:${IMAGE_TYPE}-${GITHUB_REF##*/}"
fi
```

scripts/test
colekettler: Tracking in #1244.

rastervision_pytorch_backend/rastervision/pytorch_backend/examples/chip_classification/spacenet_rio.py
AdeelH: Fixed in #1245.

.github/workflows/release.yml
colekettler: I think that's an excellent approach and I appreciate you taking the time to write it out so I can see how it looks! That completely addresses my concerns about too much nested conditional logic, reads well, and is more in line with our typical approach. I'll work this into `cipublish`.

rastervision_pytorch_backend/rastervision/pytorch_backend/examples/chip_classification/spacenet_rio.py
colekettler: Thanks @AdeelH! I'll drop this commit.

Issue #1220, opened 2021-07-24
Semantic segmentation integration tests are failing


colekettler:
The two semantic segmentation integration tests [here](https://github.com/azavea/raster-vision/blob/d9d114054095ab837c29925f4db2e2d6ba62013e/integration_tests/integration_tests.py#L48-L61) are failing.

```
------
Test: semantic_segmentation.basic
{'kwargs': {'nochip': True},
Message: raised an exception while running
 'module': 'integration_tests.semantic_segmentation',
Details: Traceback (most recent call last):
 'task': 'semantic_segmentation'}
  File "/opt/src/integration_tests/integration_tests.py", line 269, in run_test
    _run_pipeline(pipeline_cfg, runner, tmp_dir)
  File "/opt/src/rastervision_pipeline/rastervision/pipeline/cli.py", line 129, in _run_pipeline
    pipeline_run_name=pipeline_run_name)
  File "/opt/src/rastervision_pipeline/rastervision/pipeline/runner/inprocess_runner.py", line 25, in run
    _run_command(cfg_json_uri, command, 0, 1)
  File "/opt/src/rastervision_pipeline/rastervision/pipeline/cli.py", line 217, in _run_command
    command_fn()
  File "/opt/src/rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation.py", line 132, in chip
    return super().chip(*args, **kwargs)
  File "/opt/src/rastervision_core/rastervision/core/rv_pipeline/rv_pipeline.py", line 127, in chip
    chip_scene(s.build(class_cfg, self.tmp_dir), TRAIN)
  File "/opt/src/rastervision_core/rastervision/core/data/scene_config.py", line 58, in build
    if self.label_source is not None else None)
  File "/opt/src/rastervision_core/rastervision/core/data/label_source/semantic_segmentation_label_source_config.py", line 31, in build
    rgb_class_config=self.rgb_class_config)
  File "/opt/src/rastervision_core/rastervision/core/data/label_source/semantic_segmentation_label_source.py", line 53, in __init__
    rgb_class_config)
  File "/opt/src/rastervision_core/rastervision/core/data/label_source/segmentation_class_transformer.py", line 13, in __init__
    null_class_id = class_config.get_null_class_id()
  File "/opt/src/rastervision_core/rastervision/core/data/class_config.py", line 35, in get_null_class_id
    raise ValueError('null_class is not set')
ValueError: null_class is not set
```

```
------
[...]

PR #1225, opened 2021-07-29
Make RandomWindowGeoDataset more efficient when sampling chips from scenes with AOIs

### LINKED ISSUE #1219: RandomWindowGeoDataset should be more efficient when sampling chips from a scene with AOIs
Currently, `RandomWindowGeoDataset` does a kind of rejection sampling, where it samples chips randomly from anywhere in the scene and then checks to see if they fall within the AOIs, and if not, it retries a fixed number of times and then gives up:
https://github.com/azavea/raster-vision/blob/master/rastervision_pytorch_learner/rastervision/pytorch_learner/dataset/dataset.py#L338-L357

This is very inefficient when the AOIs are scattered throughout the scene and make up only a small proportion of the total area.

Instead, it could just sample from within the AOI. In general, i.e for arbitrary AOIs, this is a non-trivial problem. Some starting points:
- https://mathematica.stackexchange.com/questions/57938/how-to-generate-random-points-in-a-region
- https://gis.stackexchange.com/questions/6412/generate-points-that-lie-inside-polygon
- http://kuanbutts.com/2017/08/17/delaunay-triangulation/

AdeelH:
## Overview

Makes the sampling of random windows from scenes with AOIs more efficient. The previous approach was to sample a random window from anywhere within the scene and check to see if it falls within the AOI. In contrast, this approach is able to generate random window locations that are always inside the AOI. We still need to check if the window is fully contained within the AOI, since sampled windows near the boundaries of the AOI will occasionally overflow the AOI.

The algorithm can be summarized as:
- Partition the AOI polygons into triangles (triangulation).
    - This procedure is also able to account for holes inside polygons (see example below).
- When sampling, pick a triangle (`ABC`) at random with probability proportional to its area.
- Starting at `A`, travel a random distance along vectors `AB` and `AC`.
- Return the final position.

#### Example AOI
aoi

#### Old sampling approach
random_chips

##### triangulation
triangulation
##### sampled windows
aoi_chips

### Checklist

### Notes

- New dependency added: [triangle](https://github.com/drufat/triangle)

## Testing Instructions
- Rebuild docker image to install new dependency.
- Run an example that has an AOI, while using a `GeoDataWindowConfig` with `method=GeoDataWindowMethod.random`.

Closes #1219

### DIFF
--- a/rastervision_pytorch_learner/rastervision/pytorch_learner/learner_config.py
+++ b/rastervision_pytorch_learner/rastervision/pytorch_learner/learner_config.py
@@ -475,6 +475,15 @@ class GeoDataWindowConfig(Config):
         description='Max attempts when trying to find a window within the AOI '
         'of a scene. Only used if method = random and the scene has '
         'aoi_polygons specified.')
+    efficient_aoi_sampling: bool = Field(
+        True,
+        description='If the scene has AOIs, sampling windows at random '
+        'anywhere in the extent and then checking if they fall within any of '
+        'the AOIs can be very inefficient. This flag enables the use of an '
+        'alternate algorithm that only samples window locations inside the '
+        'AOIs. Only used if method = random and the scene has aoi_polygons '
+        'specified. Defaults to True',
+    )
 
     def validate_config(self):
         if self.method == GeoDataWindowMethod.sliding:
--- a/tests/pytorch_learner/dataset/test_aoi_sampler.py
+++ b/tests/pytorch_learner/dataset/test_aoi_sampler.py
@@ -1,0 +1,69 @@
+import unittest
+from itertools import product
+
+import numpy as np
+from scipy.stats import chisquare
+from shapely.geometry import Polygon, MultiPolygon, MultiPoint
+
+from rastervision.pytorch_learner.dataset.utils import AoiSampler
+
+
+class TestAoiSampler(unittest.TestCase):
+    def test_sampler(self, nsamples: int = 200):
+        """This test attempts to check if the points are distributed uniformly
+        within the AOI.
+
+        To do this, it performs the following steps:
+        - Create an AOI in the form of a plus-sign shape centered in a 6x6
+        grid.
+        - Create an AoiSampler for this AOI.
+        - Break the AOI up into small 1x1 blocks.
+        - Use the AoiSampler to sample nsamples points.
+        - Count the points that fall in each block.
+        - Perform a Chi-squared test to see how the counts compare to the
+        expected counts (= nsamples / nblocks).
+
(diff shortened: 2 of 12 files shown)


jamesmcclain:
:tada:

rastervision_pytorch_learner/rastervision/pytorch_learner/learner_config.py
```diff
@@ -467,6 +467,15 @@ class GeoDataWindowConfig(Config):
         description='Max attempts when trying to find a window within the AOI '
         'of a scene. Only used if method = random and the scene has '
         'aoi_polygons specified.')
+    efficient_aoi_samplings: bool = Field(
```
AdeelH, line 470: This is mostly here to provide a way to turn this functionality off on the off chance we encounter an input that the triangulation algorithm cannot handle.

AdeelH: Oops. The 's' at the end is a typo. This supposed to be called `efficient_aoi_sampling` not `efficient_aoi_samplings`.

tests/pytorch_learner/dataset/test_aoi_sampler.py
```diff
@@ -0,0 +1,69 @@
...
+            if b.within(hbar) or b.within(vbar):
+                blocks.append(b)
+        blocks = MultiPolygon(blocks)
+
+        points = MultiPoint(aoi_sampler.sample(n=nsamples))
+        # number of points in each block
+        counts = np.array(
+            [len(block.intersection(points)) for block in blocks])
+
+        p_value = chisquare(counts).pvalue
```
lewfish, line 62: Nice use of a hypothesis test.

lewfish (approved):
Nice addition! This is approved assuming you manually ran the style tests, unit tests, and integration tests locally, since CI is currently not running (but see https://github.com/azavea/raster-vision/issues/1220) Note that the semseg integration tests are failing https://github.com/azavea/raster-vision/issues/1220

AdeelH:
@lewfish One important change since the review:

I've reverted the commit that applied the inverse affine transform in the GeoDataset. The reason that was needed in the first place was that `RasterioCRSTransformer.from_dataset` used to return an `IdentityCRSTransformer` when the `.tif` did not have a CRS even if it did have an affine transform (which was the case with the other project's data). Once I modified `RasterioCRSTransformer.from_dataset` to handle the case of no-CRS + non-identity-transform, everything else just started working as expected.

I'm okay with waiting till the CI is fixed before merging this.

lewfish:
Can you do something to trigger CI on this PR? It looks like you force pushed yesterday but I'm not sure if that was after the CI PR was merged.

Issue #962, opened 2020-07-13
Train from STAC DatasetConfig


echeipesh:
In order to train a user user must provide a `DatasetConfig` ex:

https://github.com/azavea/raster-vision/blob/c63b2ce21d20b8055fed4cd6595ab6a807523fdf/rastervision_pytorch_backend/rastervision/pytorch_backend/examples/tiny_spacenet.py#L52-L59

This specifies:
- Source of training scenes
- Source of validation scenes
- Per scene label file location
- per scene raster location

In addition a `ClassConfig` is required ex:

https://github.com/azavea/raster-vision/blob/c63b2ce21d20b8055fed4cd6595ab6a807523fdf/rastervision_pytorch_backend/rastervision/pytorch_backend/examples/tiny_spacenet.py#L23-L24

One possible source of above information is from a STAC catalog using [label extension](https://github.com/radiantearth/stac-spec/tree/master/extensions/label)

There should be a library-level feature that conveniently provides a way to train a model from such a catalog with minimum configuration. Perhaps a `StacDatasetConfig`.

Proposed scheme:

- Use STAC collections to group multiple STAC label items
- Specify which STAC collection to use for training data
- Read `ClassConfig` from STAC collection JSON
- Use each label item to create label source
  - Each label item links to either GeoJSON labels or raster labels
- Use each label item to create raster source
  - Each label item links to source imagery to which labels apply

lewfish:
> * Specify which STAC collection to use for training data

To clarify, you mean that there will be a STAC collection for the training scenes, and a separate STAC collection for validation scenes?

lewfish:
As I mentioned the other day, a simple way to implement this might be to have a `to_dataset_config()` method that parses the underlying STACs and returns a standard `DatasetConfig` that links to standard `SceneConfig` objects. After that conversion happens, then everything else in Raster Vision can be the same.

echeipesh:
I started by creating a sample STAC catalog we can train from to think through the data formatting issues.

See https://gist.github.com/echeipesh/2d2a18b59d634ecbfd97b7d32bba6164
Note that `image-` and `label-` prefix in file names indicate that they would be in `image/` and `label/` subfolders if gist supported sub-folders. All the links are built as if those folders exist.

Separating the images and labels into their own collections is good and very uncontroversial.
There is some discussion on how to achieve the training and testing split.

[...]

PR #1243, opened 2021-08-17
Update STAC-reading functionality to make it compatible with STAC v1.0.*

### LINKED ISSUE #1238: Add support for parsing STAC inputs
This has been discussed before in #873 and #962.

It seems like the simplest way would be to add helper functions for reading from STAC catalogs that can be used in the `get_config()` to extract images, AOIs, etc. We have been doing this on other projects, so it should be a simple matter of adding those functions to the repo.

AdeelH:
## Overview

This PR
- fixes `parse_stac()` in `rastervision.core.utils.stac`
- adds a `read_stac()` util function to `rastervision.pytorch_backend.examples.utils` that uses `parse_stac()`
- adds a unit test for `read_stac()`

### Checklist

Closes #1238 
Connects #962


rastervision_core/rastervision/core/utils/stac.py
```diff
@@ -28,32 +32,73 @@ def my_write_method(uri, txt):
...
-
-    if not labels_uri:
-        raise ValueError('Unable to read labels URI from STAC.')
-    if not geotiff_uris:
-        raise ValueError('Unable to read GeoTIFF URIs from STAC.')
-    return labels_uri, labels_box, geotiff_uris
+
+    label_items = [
+        item for item in cat.get_all_items() if 'label' in item.stac_extensions
+    ]
```
jisantuc, line 56: This should use the `label` extension URI, since these shorthands are formally speaking not allowed anymore. I'll update the STAC exports in RF to do the same.

That url is at: https://raw.githubusercontent.com/stac-extensions/label/v1.0.0/json-schema/schema.json

That was changed in [1.0.0-rc.4](https://github.com/radiantearth/stac-spec/blob/master/CHANGELOG.md#v100-rc4---2021-05-11)

rastervision_core/rastervision/core/utils/stac.py
```diff
@@ -28,32 +32,73 @@ def my_write_method(uri, txt):
...
+
+    if len(label_items) == 0:
+        raise ValueError('Unable to find any label items in STAC catalog.')
+
+    if len(label_items) > 1:
+        raise ValueError('Found more than one label item in STAC catalog.')
```
jisantuc, line 62: The exports from GroundWork will have plural label items when there's more than one project in the GroundWork campaign, so this will be pretty brittle. It might be useful for this instead for this function to take a label item ID and try to find the label item with that ID instead of relying on singularity.

jisantuc (changes requested):

rastervision_core/rastervision/core/utils/stac.py
AdeelH: You mean that label items will have a `uri` field containing the URL `"https://raw.githubusercontent.com/stac-extensions/label/v1.0.0/json-schema/schema.json"`?

jisantuc: That's correct, or, you know, they will after raster-foundry/raster-foundry#5614

jisantuc: Another note -- yesterday I released a bugfix on the label extension, so there's now also a [v1.0.1](https://stac-extensions.github.io/label/v1.0.1/schema.json). Also, these can be addressed either by the `stac-extensions.github.io` address or the `raw.githubusercontent.com` address, since the STAC rules at this point are to resolve the schema and validate rather than convert the shorthand (I don't think consumers who want to identify items of a certain type were a big part of that conversation, though I wasn't really in it so I can't say that for certain).

AdeelH: Is resolving the schema and checking that title == "Label Extension" good enough?

rastervision_core/rastervision/core/utils/stac.py
AdeelH: Is it correct to associate an image with a label by resolving the label item's links that have rel == "source"? Can label items have more than one such link?

jisantuc: Yes that's correct. Having multiple sources [isn't explicitly forbidden](https://github.com/stac-extensions/label/#links-source-imagery) but we don't do that and it's not clear how to interpret that case to me

rastervision_core/rastervision/core/utils/stac.py
jisantuc: Oh that's a smart way to do it 👍🏻

rastervision_pytorch_backend/rastervision/pytorch_backend/examples/utils.py
```diff
@@ -139,3 +142,39 @@ def p2m(x, y, z=None):
...
+    Args:
+        uri (str): Either a URI to a STAC catalog JSON file or a URI to a zip
+            file containing a STAC catalog JSON file.
+
+    Raises:
+        FileNotFoundError: If catalog.json is not found inside the zip file.
+        Exception: If multiple catalog.json's are found inside the zip file.
+
+    Returns:
+        dict: A dict containing the extracted contents.
```
lewfish, line 161: Consider documenting the shape of the dict

tests/pytorch_backend/examples/test_utils.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class TestUtils(unittest.TestCase):
+    def test_read_stac(self):
+        zip_path = data_file_path('catalog.zip')
+        out = read_stac(zip_path)
+        self.assertIsInstance(out, list)
+        self.assertEqual(len(out), 1)
+        self.assertIsInstance(out[0], dict)
+        self.assertEqual(len(out[0]['image_uris']), 1)
```
lewfish, line 15: Consider asserting that the labels are there too

lewfish (changes requested):

lewfish (approved):

PR #1308, opened 2021-11-26
Allow MultiRasterSource to read from sub raster sources with non-identical extents and resolutions

AdeelH:
## Overview

This PR modifies `MultiRasterSource` so that it is now able to read from sub raster sources that do not necessarily have the same extents or resolutions. This is achieved via the following algorithm:
```
- using a window in pixel coords, get a chip from the first sub raster source
- convert the window to world coords using the CRS of the first sub raster source
- for each of the remaining sub raster sources
  - convert the world-coords window into pixel coords using the sub raster source's CRS
  - get a chip from the raster source using this window, specify `out_shape` when reading to ensure the shape matches the first chip
- stack chips together into a single image array
```

The PR also removes the `allow_different_extents` field (#1004) from `MultiRasterSourceConfig`.

### Checklist


rastervision_core/rastervision/core/data/raster_source/multi_raster_source.py
```diff
@@ -148,11 +167,26 @@ def get_chip(self, window: Box) -> np.ndarray:
         Returns:
             np.ndarray with shape [height, width, channels]
         """
-        chip_slices = [rs.get_chip(window) for rs in self.raster_sources]
+        if self.all_extents_equal:
```
lewfish, line 161: The code in `get_chip` and `_get_chip` are almost identical. Consider extracting the common code to a function.

AdeelH: Good point. Fixed.

PR #1337, opened 2022-02-08
Add support for external models for object detection

### LINKED ISSUE #1256: Extend support for external models and loss functions to the object detection task
## 🚀 Feature

Support for external models and loss functions to the object detection task.

## Motivation

- #985 and #992 added the ability to use external model and loss function definitions for all tasks except for object detection because of its slightly more complex nature of the existing model architecture. However, it should still be possible for the user to use a different architecture if its inputs and outputs are compatible with the `ObjectDetectionLearner`.
- External model and loss function support is a significant RV feature and It would be nice to no longer have the not-for-OD caveat attached to it.

## Pitch

The `ObjectDetectionLearner` should make use of the external model and loss function configs if provided.

## Alternatives

N/A

## Additional context

Incorporating an external loss function might be tricky given that currently the loss in computed inside the model class.

AdeelH:
## Overview

This PR
- Renames the `MyFasterRCNN` adapter to `TorchVisionODAdapter`.
- Moves backbone-creation code out of the adapter and into `ObjectDetectionLearner.build_model()`.
- Refactors `BoxList`. See commit body for details.
- Abstracts out the conversions to/from `BoxLists` into their own functions for clarity and easier testing.
- Adds unit tests for the new adapter and `BoxList`.

As a workaround for limitations of older versions of TorchVision, the adapter code previously used to add 2 extra classes to model. In newer TorchVision versions, only one such extra class is needed. However, to maintain backward compatibility with models trained with earlier Raster Vision versions, `build_model()` still adds 2 extra classes. External models, however, are not subject to this backward compatibility constraint and only need to add one extra class. The code assumes this to be true when dealing with external models.

Also updates the plotting function to make use of util functions provided by TorchVision. Results look like this:
image

## Comparison with previous Faster RCNN implementation:
- Small speed up in average epoch training time (mm:ss): `01:47 --> 01:29`.
- No speed up in average validation time (mm:ss): `00:15 --> 00:16`.
- Slight, probably statistically insignificant, improvement in metrics:
```sh
python "rastervision_pytorch_backend/rastervision/pytorch_backend/examples/test.py" \
compare \
"s3://raster-vision/examples/0.13/output/cowc-potsdam-od/" \
"s3://raster-vision-ahassan/rvtest/od/cowc/"
```
image

## Comparison between Faster RCNN and SSD:
- SSD runs much faster with an average epoch training time of `01:00` (compared to `01:29` with ResNet-18 Faster RCNN).
- Average validation time of `00:11`.
- SSD performs considerably worse:
```sh
python "rastervision_pytorch_backend/rastervision/pytorch_backend/examples/test.py" \
compare \
"s3://raster-vision-ahassan/rvtest/od/cowc/" \
"s3://raster-vision-ahassan/rvtest/od/cowc_ssd/"
```
image

### Checklist

### Notes

## Testing Instructions

* How to test this PR
- Run `cowc_potsdam.py` with `-a external_model True`.

Closes #1256

### DIFF
--- a/rastervision_pytorch_backend/rastervision/pytorch_backend/examples/object_detection/cowc_potsdam.py
+++ b/rastervision_pytorch_backend/rastervision/pytorch_backend/examples/object_detection/cowc_potsdam.py
@@ -24,6 +24,7 @@ def get_config(runner,
                processed_uri: str,
                root_uri: str,
                nochip: bool = True,
+               external_model: bool = False,
                test: bool = False) -> ObjectDetectionConfig:
     """Generate the pipeline config for this task. This function will be called
     by RV, with arguments from the command line, when this example is run.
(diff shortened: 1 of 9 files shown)


jamesmcclain:
:eyes:

lewfish (approved):

AdeelH:
Small post-approval edit: renamed variable from `class_colors` to `box_colors` in `plot_xyz()`.

PR #1345, opened 2022-02-15
Extend multiband support to all tasks

### LINKED ISSUE #1003: Multi-Raster Source is Hard-Coded Not to Use Transformers (In Some Places)

### LINKED ISSUE #1255: Extend multiband support to all tasks
## 🚀 Feature

Support for multiband rasters for all tasks.

## Motivation

- #972 and #1026 added the ability to use rasters with arbitrary number of bands for semantic segmentation, but there is no good reason why this functionality should remain limited to semantic segmentation only.
- Multiband support is a significant RV feature and It would be nice to no longer have the only-for-SS caveat attached to it.

## Pitch

The other tasks should implement support for the following:
- Modification to the input channels of the first conv layer of the backbone CNN.
- `channel_display_groups`.
- Saving chips as numpy files instead of png files.

## Alternatives

N/A

## Additional context

N/A

### LINKED ISSUE #1258: channel_display_groups in SemanticSegmentationConfig is not used
## 🐛 Bug

`SemanticSegmentationConfig` does not make use of its `channel_display_groups` field. 

## To Reproduce

1. Set `channel_display_groups` in `SemanticSegmentationConfig` but not in the data config.
2. Inspect the value of `channel_display_groups` in the learner config. 
3. It should be the default value (`channel_display_groups={'Input': (0, 1, 2, 3)}`) and not the value set in the `SemanticSegmentationConfig`.

## Expected behavior

The field should be removed from `SemanticSegmentationConfig` and all the validation checks should be moved to `SemanticSegmentationDataConfig`. Will require some backward compatibility code.

## Environment

N/A

## Additional context

This bug was introduced here: https://github.com/azavea/raster-vision/commit/276f0ddd5089ea721fc51aa205d45367c5b2a695#diff-d66ddf094acc66710795e7ad65820793ce36deef5fa1dd7b1039097f059fcc06L35.

Prior to #1046, it was passed to `SemanticSegmentationDataConfig` in the backend config, but #1046 made the data config, and this field within it, directly specifiable.

### LINKED ISSUE #1321: The Argument `img_channels` is required in `SemanticSegmentationImageDataConfig`, but in a Non-Obvious Way.
## 🐛 Bug

The argument `img_channels` is required in `SemanticSegmentationImageDataConfig`, but in a non-obvious way.

## To Reproduce

I am try to create a pipeline to create chips from AVIRIS imagery, but I do not need it to do any work aside from that.  My declaration of the semantic segmenation data config object looks like this
```python
    data = SemanticSegmentationImageDataConfig(
        # img_channels=224,
        img_sz=chip_sz,
        num_workers=0,
        preview_batch_limit=8
    )
```
but I get an error saying that `img_channels` cannot be `None` regardless of whether the `img_channels` argument is directly supplied or not.  The error follows
```
 File "pydantic/main.py", line 274, in pydantic.main.BaseModel.__init__ 
pydantic.error_wrappers.ValidationError: 1 validation error for SemanticSegmentationImageDataConfig
img_channels
  none is not an allowed value (type=type_error.none.not_allowed)
```

The only way to circumvent this that I found was to supply an appropriately-sized list of channels to the raster source elsewhere in the pipeline
```python
    image_source = RasterioSourceConfig(
        channel_order=list(range(224)),
        uris=[imagery],
        allow_streaming=True,
        transformers=transformers,
        extent_crop=extent_crop,
    )
```

## Expected behavior

I would (possibly) expect that supplying `num_channels` to the data config would be sufficient, or alternatively the error message could be different.

## Environment

A docker container made from current (at time of writing) `master`.

AdeelH:
## Overview

Extends functionality introduced in #972 to other tasks. See commit bodies for more details.

In summary, this PR
- allows all learners to modify the model's backbone's first conv layer (as long as it is a ResNet backbone) to be able to take in arbitrarily many channels
- allows all pytorch backends to write non-3-channel chips as numpy array if using the chip stage
- moves channel related configurations from SS classes to parent classes
- moves some channel related configurations from core classes to learner classes
- does away with `img_format` and `label_format` configurations
  - all ImageDataset classes can now read both
- allows all learner classes to plot non-3-channel images using `channel_display_groups` (which now lives in `PlotOptions`)
  - also improves visualization of classification and regression ground truth and predictions
- adds upgraders to handle backward-compatibility for changed `Config`s
- Fixes #1258
- Seems to fix #1321
- Fixes #1003
  - `get_img_channels()` which has been moved from SS Pipeline to `PytorchLearnerBackendConfig` now uses `use_transformers=True` if it needs to build a scene.

Other changes
- makes all ImageDataset classes more similar to each other
- adds a bunch of unit tests
  - notably, for all learner classes resulting in a big jump in test coverage
- adds a progress bar to the chip stage
- fixes some warnings

### Object detection
valid
test_preds

### Checklist

### Notes

N/A

## Testing Instructions

* How to test this PR
- use the `multiband` flag in `cowc_potsdam.py`

Closes #1255 
Fixes #1258
Fixes #1321 
Fixes #1003


lewfish (approved):
Excellent PR!

rastervision_pytorch_learner/rastervision/pytorch_learner/learner_config.py
```diff
@@ -323,12 +379,14 @@ class DataConfig(Config):
...
         if not self.class_colors:
             self.class_colors = [color_to_triple() for _ in self.class_names]
+        self.plot_options.update(self)
 
     def validate_augmentors(self):
         self.validate_list('augmentors', augmentors)
 
     def validate_config(self):
         self.validate_augmentors()
+        self.validate_augmentors()
```
AdeelH, line 389: TODO: undo accidental line duplication

rastervision_core/rastervision/core/data/raster_source/rasterio_source.py
```diff
@@ -242,8 +242,9 @@ def get_chip(self, window,
 
     def _activate(self):
         # Download images to temporary directory and delete them when done.
-        self.image_tmp_dir = tempfile.TemporaryDirectory(dir=self.tmp_dir)
-        self.imagery_path = self._download_data(self.image_tmp_dir.name)
+        if self.image_tmp_dir is None:
+            self.image_tmp_dir = TemporaryDirectory(dir=self.tmp_dir)
+        self.imagery_path = self._download_data(self.image_tmp_dir)
```
AdeelH, line 247: TODO: `self.image_tmp_dir` --> `self.image_tmp_dir.name`

PR #1375, opened 2022-04-01
Provide grouping mechanism for scenes and use it in the analyze and eval stages

### LINKED ISSUE #984: Provide grouping mechanism for evaluation
Currently the validation scenes are each evaluated individually and then averaged together. We would like to be able to specify groups of scenes, and compute average metrics over each group. For example, this would allow us to compute the score for rural vs. urban scenes. Using the below config, you would get an `eval.json` with entries for urban and rural in addition to averaged over all scenes.

```python
DatasetConfig(
   train_scenes=..., validation_scenes=[SceneConfig(id='nyc'), SceneConfig(id='farmland')],test_scenes=[], 
   scene_groups={
      'urban': ['ny',],
      'rural: ['farmland']
   })
```

In addition, two scene groups should be added automatically: `validation` and `test` based on the scene ids within `validation_scenes`, and `test_scenes`. This way, we can compute separate metrics for the validation and test set. Note that currently the test set is not used by the eval command at all, and is only used by predict.

AdeelH:
## Overview

This PR adds a `scene_group` field to `DatasetConfig` which always has at least 3 members: `"train_scenes"`, `"validation_scenes"`, and `"test_scenes"`.

`analyze` and `eval` commands now work at the scene group level. Each `Analyzer` and `Evaluator` is now built and run separately for each scene group. The outputs look like this:
- `<analyze_uri>/stats/<scene-group-name>/stats.json`
- `<eval_uri>/<scene-group-name>/eval.json`

`bundle` now copies the entire `analyze_uri` directory into the bundle.

### Checklist

Closes #984


lewfish (commented):
How does the `predict` command (ie. the CLI command) know which statistics to use if there's a `StatsTransformer`? (I see that the stats are written to a separate file for each group.)

AdeelH:
`StatsTransformerConfig` has a new `scene_group` field that it uses to generate its `stats_uri`.

AdeelH:
Good catch. I have added a `--scene-group` option now to deal with this.

lewfish (approved):

Issue #1395, opened 2022-04-28
Make Raster Vision easier to use as a library (and not just as a framework)


AdeelH:
## Problem
Raster Vision (RV) is a [framework](https://en.wikipedia.org/wiki/Software_framework). To use it, users must configure the RV pipeline by defining a `get_config()` function. This has the following disadvantages:
- Even the [simplest examples](https://docs.rastervision.io/en/0.13/quickstart.html#configuring-a-semantic-segmentation-pipeline) of these `get_config()` functions tend to be dozens of lines long.
- It is not obvious how to map user requirements to a suitable pipeline config without deep familiarity with RV internals.
- It is harder to debug because one cannot individually examine the various components of the pipeline to ensure that they are working as intended.
- Overriding RV default functionality requires creating [a new RV project](https://docs.rastervision.io/en/0.13/bootstrap.html?highlight=cookiecutter) which is a non-trivial amount of effort.

This high barrier-to-entry is likely discouraging for new users.

## Proposed solution
To remedy this, it is proposed that we refactor RV so that it is more easily usable as a library with multiple entry and exit points. This is in line with the general software engineering advice about [writing libraries instead of frameworks](https://www.brandons.me/blog/libraries-not-frameworks) as well as the [trajectory of other similar frameworks](https://ai.facebook.com/blog/reengineering-facebook-ais-deep-learning-platforms-for-interoperability/). 

A largely successful attempt to use RV (in its current state) as a library was made in [the companion notebook](https://colab.research.google.com/drive/1Ex5xbAx4epzReCQDsqvZLoTpF6AtfnHr?usp=sharing) to the [Change Detection blog](https://www.azavea.com/blog/2022/04/18/change-detection-with-raster-vision/).

An example of what a future workflow might look like:
![image](https://user-images.githubusercontent.com/13014700/165721385-19d6249d-d21e-4768-a7c3-17077f0f69d7.png)

## Tasks
To achieve this, we would need to complete the following tasks. Each of these should ideally be turned into its own issue.
- Ensure that RV is installable and usable outside of the Docker container.
- Ensure that classes can function without relying on global state (such as the global `tmp_dir`).
- ~Ensure that classes do not rely on Pydantic configs.~
- Simplify the initialization of a `GeoDataset` as much as possible. Currently, it requires first creating a `RasterSource`, a `LabelSource`, and a `Scene`. Can this be simplified? Is `Scene` a useful abstraction?
- Refactor `Learner`. 
[...]

PR #1413, opened 2022-05-27
Refactor `Learner` and related configs to be more flexible and easier to use in a notebook

AdeelH:
## Overview

This PR makes the `Learner` more flexible and easier to use in a notebook. It also makes it possible to build datasets, model, loss, optimizer, and schedulers from configs directly outside of a `Learner`. This is particularly useful for examining/debugging datasets before getting started with training.

Major changes:
- Moves the dataset, model, loss, optimizer, schedulers creation code from `Learner` to the respective configs' `.build()` methods.
  - The `Learner`'s `build_*` methods are still there but now they just call `.build()` on the relevant config.
- It is now possible to pass any of or all of datasets, model, loss, optimizer, schedulers to the `Learner`'s constructor, which causes the `Learner` to skip building those specific objects.
- Since it is annoying to have to manually call `.update()` and `.validate_config()` on every config, these have been removed and the code has been moved to pydantic validators that get called automatically on initialization. See commit body for more details.
  - This refactoring has currently only been applied to configs in `rastervision_pytorch_learner`, but might make sense to apply to the rest too.
- `SolverConfig.ignore_last_class` has been replaced with the more general and consistent-with-pytorch `SolverConfig.ignore_class_index`.

Minor changes:
- Added a `show` flag to `Learner.plot_batch()` and `Learner.plot_predictions()` that causes the plot to be rendered to the frontend in addition to being saved as a file.
- Replaced `click` progress bars with `tqdm` progress bars.
- `LearnerConfig.model` can now be `None`.
- Updated unit tests.

### Checklist

## Testing Instructions

See examples in this Colab notebook: https://colab.research.google.com/drive/1ybXLCNlbGRM9vAjI-FPj0S_lvz2wh_UR?usp=sharing

Connects to #1395


lewfish (approved):
I didn't review every line of code due to the size of the PR, but the high-level changes smell right to me.

PR #1440, opened 2022-08-01
Fix problems with readthedocs builds

AdeelH:
## Overview

This PR
- updates `.readthedocs.yml` to conform to the latest standard
- updates `.readthedocs.yml` to install RV plugins from source instead of from pypi
- changes version strings to 0.13.1 everywhere in RV
- makes version numbers for each dependency consistent across all RV plugins

### Checklist

- [x] Added `needs-backport` label if PR is bug fix that applies to previous minor release
- [x] Ran scripts/format_code and committed any changes
- [x] Documentation updated if needed
- [x] PR has a name that won't get you publicly shamed for vagueness

### Notes

See also the discussion here: https://github.com/azavea/raster-vision/pull/1146

## Testing Instructions

- Check if the CI doc build succeeds.

PR #1441, opened 2022-08-01
Fix a bug introduced in #1413

AdeelH:
## Overview

This PR fixes a bug introduced in #1413 that cause the weight from the model bundle to not be loaded.

### Checklist

### Notes

This was not caught by the integration tests because they expect the `init_weights` to be the same as the trained weights.

## Testing Instructions
```sh
rastervision predict \
    "s3://azavea-research-public-data/raster-vision/examples/model-zoo-0.13/isprs-potsdam-ss/model-bundle.zip" \
    "s3://raster-vision-ahassan/potsdam/data/raw/4_Ortho_RGBIR/top_potsdam_6_8_RGBIR.tif" \
    "./pred_ss_potsdam/" \
    --channel-order 3 0 1
```

The above should produce random predictions when run without this fix and good predictions when run with this fix.

### DIFF
--- a/rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
+++ b/rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
@@ -320,7 +320,7 @@ class Learner(ABC):
         self.model = self.build_model(model_def_path=model_def_path)
 
         self.model.to(self.device)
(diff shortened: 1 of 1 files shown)


AdeelH:
To be rebased on #1440

PR #1450, opened 2022-08-02
Allow caching downloads

### LINKED ISSUE #1449: Avoid downloading the same file multiple times in the same run
## 🚀 Feature

Raster Vision should not download the same file multiple times during the same run.

## Motivation

Currently, due to the `ActivateMixin` mechanism, `RasterioSource` will download then delete the raster image during its initialization and then download it again when it is used. If there are multiple raster sources that use the same raster image file, this whole thing will happen for each of them.

Example:
image

## Pitch

All files downloaded by Raster Vision should be cached. This should apply to all `FileSystem`s. Optionally, the `ActivateMixin` mechanism should be removed.

## Alternatives

Instead of caching we could just modify `RasterioSource` so that it doesn't activate and deactivate during its initialization, but then we would have to figure out the `channel_order` and `dtype` some other way.

## Additional context
Implementation options:
- We can use just use `~/.cache/raster-vision`. Before downloading, `FileSystem` can check whether the file already exists there.
- Something like the [requests-cache](https://requests-cache.readthedocs.io/en/latest/) library would work for HTTP downloads, but not for S3 downloads.

Besides implementing the caching we should reevaluate the usefulness of the `ActivateMixin` mechanism and remove it if it is not strictly necessary.

AdeelH:
## Overview

This PR makes it so that `download_if_needed()` downloads files to `{rv_config.tmp_dir_root}/cache` by default if a `download_dir` is not specified.

This allows caching of downloads by omitting `download_dir` where it is a temporary directory and we might need to download the same file multiple times.

### Checklist

### Notes

N/A

## Testing Instructions

- Open the logs for the CI build for this PR and search (using the search bar) for "Using cached". Or see screenshots below:
image
image

- Run a predict command using a remote image (example below) and check whether it downloads only once.
```sh
rastervision predict \
    "s3://azavea-research-public-data/raster-vision/examples/model-zoo-0.13/isprs-potsdam-ss/model-bundle.zip" \
    "s3://raster-vision-ahassan/potsdam/data/raw/4_Ortho_RGBIR/top_potsdam_6_8_RGBIR.tif" \
    "./pred_ss_potsdam/" \
    --channel-order 3 0 1
```

Closes #1449

### DIFF
--- a/rastervision_pytorch_learner/rastervision/pytorch_learner/utils/torch_hub.py
+++ b/rastervision_pytorch_learner/rastervision/pytorch_learner/utils/torch_hub.py
@@ -1,4 +1,3 @@
-from tempfile import TemporaryDirectory
 from typing import Any, Optional
 from pathlib import Path
 from os.path import join, isdir, realpath
@@ -7,6 +6,7 @@ from glob import glob
 
 import torch.hub
 
+from rastervision.pipeline import rv_config
 from rastervision.pipeline.file_system import (download_if_needed, unzip)
 
 
@@ -119,7 +119,7 @@ def torch_hub_load_uri(uri: str, hubconf_dir: str, entrypoint: str, *args,
     if is_zip:
         # unzip
         zip_path = download_if_needed(uri)
-        with TemporaryDirectory() as tmp_dir:
+        with rv_config.get_tmp_dir() as tmp_dir:
             unzip_dir = join(tmp_dir, uri_path.stem)
             _remove_dir(unzip_dir)
             unzip(zip_path, target_dir=unzip_dir)
--- a/rastervision_pytorch_learner/rastervision/pytorch_learner/utils/torch_hub.py
+++ b/rastervision_pytorch_learner/rastervision/pytorch_learner/utils/torch_hub.py
@@ -62,11 +62,7 @@ def get_hubconf_dir_from_cfg(cfg, parent: Optional[str] = '') -> str:
     return path
 
 
-def torch_hub_load_github(repo: str,
-                          hubconf_dir: str,
-                          entrypoint: str,
-                          *args,
-                          tmp_dir: Optional[str] = None,
+def torch_hub_load_github(repo: str, hubconf_dir: str, entrypoint: str, *args,
                           **kwargs) -> Any:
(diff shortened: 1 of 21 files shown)


AdeelH:
@lewfish, this PR isn't complete yet, but can you take a quick look at and check whether it looks directionally correct?

rastervision_pytorch_learner/rastervision/pytorch_learner/utils/torch_hub.py
```diff
@@ -136,22 +118,21 @@ def torch_hub_load_uri(uri: str, hubconf_dir: str, entrypoint: str,
...
-        # if the top level only contains a directory
-        if (len(unzipped_contents) == 1) and isdir(unzipped_contents[0]):
-            sub_dir = unzipped_contents[0]
-            shutil.move(sub_dir, hubconf_dir)
-        else:
-            shutil.move(unzip_dir, hubconf_dir)
-
-        _remove_dir(unzip_dir)
+        zip_path = download_if_needed(uri)
+        with TemporaryDirectory() as tmp_dir:
```
lewfish, line 122: I think this won't respect the temporary directory that is set in `RVConfig`. Should you use `RVConfig.get_temp_dir()`?

AdeelH: This is just for initially extracting the zip. The extracted contents are eventually moved to `hubconf_dir` (line 133 and 135) and this `tmp_dir` is destroyed immediately after. So it doesn't really where the `tmp_dir` is created.

lewfish: It could matter if the zip file is huge and extracting it in the system provided temp dir will cause it to run out of disk space. The Docker container file system only has 22GB in it, so that's why we mount the `/opt/data` directory to the host file system which has more space, and then put the temporary directories in `/opt/data`.

AdeelH: That makes sense but there are also other places where we are not using a tmp dir rooted in `RVConfig`'s tmp dir. I guess we should change those too.

lewfish (approved):

PR #1451, opened 2022-08-03
Fix #1052

### LINKED ISSUE #1052: Error loading model using Predictor
I'm using v0.12 via docker.  I want to pre-load a model so I can do multiple predictions quickly. 

When I use the `Predictor` class to load the model similar to what is done in `rastervision/core/cli.py`:

``` python 
model_path = "./model-bundle.zip"

with rv_config.get_tmp_dir() as tmp_dir:
    model = Predictor("./model-bundle.zip", tmp_dir)
```
or if I do this:

``` python 
model_path = "./model-bundle.zip"
tmp_dir = "./model_tmp"

model = Predictor("./model-bundle.zip", tmp_dir)
```
I get the following error:

```sh
ConfigError: The plugin_versions field contains an unrecognized plugin name: rastervision.core.
```

Predictions work fine when I use the cli but when I try to use the `Predictor` class myself, I get an error.

AdeelH:
## Overview

This PR fixes the bug reported and discussed in #1052.

In short, the bug causes `rastervision.core` to not be registered in the [`Registry`](https://github.com/azavea/raster-vision/blob/master/rastervision_pipeline/rastervision/pipeline/registry.py).

A different manifestation of this error has been seen and fixed before. Specifically, in [this commit](https://github.com/azavea/raster-vision/commit/10e9cf1cf3b83f0d728446ac754bb203f047eb91) which describes it like so:
<blockquote>
This is meant to fix a problem that occurred when running a unit test [...]. Specifically, when [...] `upgrade_config(config_dict)` [was called], it would throw a `KeyError` for "rastervision.core" due to "rastervision.core" being absent from the registry.

This was traced down to be due to the following workflow:
- `unittest discover` runs all discovered test files in alphabetical order
- the first file happens to be `tests/core/data/crs_transformer/test_rasterio_crs_transformer.py`
- the line `from rastervision.core.data.crs_transformer import RasterioCRSTransformer` in this file is run
- this causes core's `__init__.py`'s `import rastervision.pipeline` line to be run
- this causes the registry to be initialized
- the registry attempts to find all the plugins and load their `register_plugin()` functions
- (not 100% certain about this part but it seems like) when the registry tries to import core, it results in a circular import. At this point, core is already in `sys.modules` but has not been fully executed yet. Which means its `register_plugin()` hasn't been defined yet! Some context: https://stackoverflow.com/a/744403/5908685 .
- the registry fails to find core's `register_plugin()` and therefore fails to register it
</blockquote>

To summarize, when a Raster Vision plugin calls `import rastervision.pipeline` in its `__init__.py` before defining a `register_plugin()` function, that plugin will fail to get registered.

This PR fixes this problem by simply moving the `register_plugin()` definition to the top in each plugin's top-level `__init__.py`.

### Checklist

### Notes

N/A

## Testing Instructions

Create a python file with the following contents (or run in a Python/IPython shell):
```python
from rastervision.core.predictor import Predictor

model_bundle_path = 's3://azavea-research-public-data/raster-vision/examples/model-zoo-0.13/isprs-potsdam-ss/model-bundle.zip'
tmp_dir = '/opt/tmp'
predictor = Predictor(model_bundle_path, tmp_dir)
```

This should fail without the changes in this PR and succeed with them.

Fixes #1052

### DIFF
--- a/cookiecutter_template/{{cookiecutter.project_name}}/rastervision_{{cookiecutter.project_name}}/rastervision/{{cookiecutter.project_name}}/__init__.py
+++ b/cookiecutter_template/{{cookiecutter.project_name}}/rastervision_{{cookiecutter.project_name}}/rastervision/{{cookiecutter.project_name}}/__init__.py
@@ -1,6 +1,9 @@
 # flake8: noqa
-import rastervision.pipeline
-import rastervision.{{cookiecutter.project_name}}.test_pipeline_config
+
 
 def register_plugin(registry):
     pass
+
+
+import rastervision.pipeline
+import rastervision.{{cookiecutter.project_name}}.test_pipeline_config
--- a/rastervision_aws_batch/rastervision/aws_batch/__init__.py
+++ b/rastervision_aws_batch/rastervision/aws_batch/__init__.py
@@ -1,10 +1,9 @@
 # flake8: noqa
 
-import rastervision.pipeline
-from rastervision.aws_batch.aws_batch_runner import *
-
 
 def register_plugin(registry):
+    from rastervision.aws_batch.aws_batch_runner import (AWS_BATCH,
+                                                         AWSBatchRunner)
     registry.set_plugin_version('rastervision.aws_batch', 0)
     registry.set_plugin_aliases('rastervision.aws_batch',
                                 ['rastervision2.aws_batch'])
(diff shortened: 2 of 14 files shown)


lewfish (approved):
Nice!

PR #1470, opened 2022-09-02
[WIP] Add example notebooks

AdeelH:
## Overview

### Checklist

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

Closes #XXX


.gitignore
```diff
@@ -112,6 +112,8 @@ deployment/ansible/*.retry
 *.aux.xml
 
 /data/
-/notebooks/
 /.vscode/
 /.history/
+
+/notebooks/*
```
lewfish, line 118: There are some places in the notebooks where you display an image with matplotlib and I think it would look better if you used `axes('off')`

.gitignore
```diff
@@ -112,6 +112,8 @@ deployment/ansible/*.retry
...
+!/notebooks/**/*.ipynb
```
lewfish, line 119: In my PR I would like to modify some parts of the notebooks to use the `Visualizer` class (when appropriate) instead of plotting stuff from scratch. As part of this I will modify the `Visualizer` to work when you just have images and no labels.

rastervision_core/rastervision/core/evaluation/semantic_segmentation_evaluation.py
```diff
@@ -56,59 +43,3 @@ def compute(self, gt_labels: 'SemanticSegmentationLabels',
             self.class_to_eval_item[class_id] = eval_item
 
         self.compute_avg()
-
-    def compute_vector(self, gt: Union[str, dict],
-                       preds: Iterable[Union[str, dict]],
-                       vector_outputs: Iterable['VectorOutputConfig']) -> None:
```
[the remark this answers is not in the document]
AdeelH: There is no current implementation in RV; RV uses [mask-to-polygons's implementation](https://github.com/azavea/mask-to-polygons/blob/master/mask_to_polygons/processing/score.py). So the question is: do we want to move it into RV and maintain it?

I am against it for the reasons below.
- It seems to be most appropriate for classes whose instances tend to be small, spatially disjoint units like cars or small buildings, but not for classes like roads or vegetation or large buildings etc. [Image 5 in this article](https://medium.com/the-downlinq/the-spacenet-metric-612183cc2ddb) makes me skeptical of its usefulness. So it is not general enough.
- It seems more appropriate for object detection than semantic segmentation and I am not sure if it is used for semantic segmentation outside of SpaceNet competitions.
- We already produce basic pixel-level metrics for semantic segmentation. Adding additional specialized metrics is a slippery slope given how many there are.
- Moving this into RV would require a rewrite to make use of geopandas to get rid of the shapely warnings.

rastervision_core/rastervision/core/data/label_store/semantic_segmentation_label_store_config.py
```diff
@@ -15,17 +15,18 @@ class VectorOutputConfig(Config):
...
+        'of a SceneConfig and RVPipeline, this field will be auto-generated.')
     class_id: int = Field(
         ...,
         description='The prediction class that is to turned into vectors.')
     denoise: int = Field(
-        0,
-        description=
-        ('Radius of the structural element used to remove high-frequency signals from '
-         'the image.'))
+        8,
```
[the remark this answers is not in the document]
AdeelH: If this is zero and the raster is large and the predictions are small and scattered, gdal's polygonization can get stuck. We have had instances of people asking about why this is happening and I've experienced this myself. From experience, a value of 8 is usually large enough to make it not get stuck, while being small enough that it does not remove useful detections. `pi * 8 * 8 ~= 200 px^2` which is only ~0.3% of a 256x256 chip.

.gitignore
AdeelH: I left those axes on to call attention to the dimensions of the rasters being plotted.

notebooks/misc.ipynb
```diff
@@ -0,0 +1,89 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
```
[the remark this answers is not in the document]
AdeelH: IDK if it's useful enough to be its own notebook. It's not even using RV. Maybe we can leave it here for now and move it to the STAC notebook when we have that.

.gitignore
AdeelH: Do you mean the notebooks in this PR? I think it makes sense for `sampling_training_data.ipynb` because it uses `pytorch_learner` but not the others since they are just using `core` and `pipeline`.

notebooks/misc.ipynb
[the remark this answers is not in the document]
AdeelH: I'm also open to removing this entirely for now.

.gitignore
lewfish: Yes, I will just make changes to `sampling_training_data` then.

lewfish (approved):
Your clarifications made sense. I think we should leave things as is.

PR #1486, opened 2022-09-29
Allow discarding edges of predicted chips in SS, add `Labels.save()` convenience method and refactor `SemanticSegmentationDiscreteLabels`

AdeelH:
## Overview

This PR
- Allow cropping edges of predicted chips in SS to reduce boundary artifacts.
  - See `SemanticSegmentationLabels.from_predictions()` and `SemanticSegmentationPredictOptions`.
  - This can be used in conjunction with existing merging functionality for even better results (see images below).
- Adds a `Labels.save()` convenience method, obviating the need to separately instantiate a `LableStore`.
- Re-implements `SemanticSegmentationDiscreteLabels` such that it stores data in an array instead of a dict. This makes it more similar to `SemanticSegmentationSmoothLabels`.
    - This array stores the counts for how many times each pixel has been classified as each class, which allows merging overlapping chips in a principled way.
    - This makes it more memory efficient if using `stride` < `chip_sz`.
    - It also makes `SemanticSegmentationEvaluation` less brittle since the gt and pred labels no longer need to have the same dict keys.

Minor changes:
- `RasterioSource` now throws warnings during initialization about sub-optimal raster tiling.
- Removed wasteful calls to `SemanticSegmentationLabelSource.enough_target_pixles()` with target pixels set to _all_ class IDs.
- Other minor refactoring and type hint improvements.
- Add/update unit tests.

**Cropping demo**
`s3://raster-vision-ahassan/rvtest/pr_1486/predict/`

ss_pred_crop_building
ss_pred_crop_trees

### Checklist

### Notes

N/A

## Testing Instructions
- See new/updated unit tests.
- Examine cropping results:
  ```py
  rs_no_overlap = RasterioSource('s3://raster-vision-ahassan/rvtest/pr_1486/predict/no_crop_2/2_12/scores.tif')
  rs_no_crop = RasterioSource('s3://raster-vision-ahassan/rvtest/pr_1486/predict/no_crop/2_12/scores.tif')
  rs_crop = RasterioSource('s3://raster-vision-ahassan/rvtest/pr_1486/predict/crop_auto/2_12/scores.tif')
  rs_crop_merge = RasterioSource('s3://raster-vision-ahassan/rvtest/pr_1486/predict/overlap_100_crop_25/2_12/scores.tif')
  ```

### DIFF
--- a/rastervision_core/rastervision/core/data/label/semantic_segmentation_labels.py
+++ b/rastervision_core/rastervision/core/data/label/semantic_segmentation_labels.py
@@ -31,8 +31,8 @@ class SemanticSegmentationLabels(Labels):
         self.ymin, self.xmin, self.width, self.height = extent.to_xywh()
         self.dtype = dtype
 
-    def _to_local_coords(self, window: Union[Box, Tuple[int, int, int, int]]
-                         ) -> Tuple[int, int, int, int]:
+    def _to_local_coords(self,
+                         window: Union[Box, Tuple[int, int, int, int]]) -> Box:
         """Convert window to extent coordinates.
 
         Args:
@@ -40,7 +40,7 @@ class SemanticSegmentationLabels(Labels):
                 Box or a 4-tuple of (ymin, xmin, ymax, xmax).
 
         Returns:
-            Tuple[int, int, int, int]: a 4-tuple of (ymin, xmin, ymax, xmax).
+            Box: Window in local coords.
         """
         ymin, xmin, ymax, xmax = window
         # convert to extent coords
@@ -52,7 +52,7 @@ class SemanticSegmentationLabels(Labels):
         # clip max values to array bounds
         ymax_local = min(self.height, ymax_local)
         xmax_local = min(self.width, xmax_local)
-        return ymin_local, xmin_local, ymax_local, xmax_local
+        return Box(ymin_local, xmin_local, ymax_local, xmax_local)
 
     @abstractmethod
     def __add__(self, other) -> 'SemanticSegmentationLabels':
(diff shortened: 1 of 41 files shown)


rastervision_core/rastervision/core/data/label/semantic_segmentation_labels.py
```diff
@@ -200,12 +202,22 @@ def from_predictions(cls,
             extent (Box): The extent of the region to which the labels belong,
                 in global coordinates.
             num_classes (int): Number of classes.
+            crop_sz (Optional[int]): Number of rows/columns of pixels from the
```
lewfish, line 205: If `crop_sz` is > 0, then it seems like the first window should start at `(-crop_sz, -crop_sz)` instead of `(0, 0)`.

rastervision_core/rastervision/core/data/label/semantic_segmentation_labels.py
```diff
@@ -185,9 +186,10 @@ def make_empty(cls, extent: Box, num_classes: int, smooth: bool = False
     def from_predictions(cls,
                          windows: Iterable['Box'],
                          predictions: Iterable[Any],
-                         extent: Optional[Box] = None,
-                         num_classes: Optional[int] = None,
-                         smooth: bool = False
+                         extent: Box,
+                         num_classes: int,
+                         smooth: bool = False,
+                         crop_sz: Optional[int] = None
```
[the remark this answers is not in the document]
AdeelH: The `'auto'` thing is just a convenience that `SemanticSegmentationPredictOptions` provides. I'd rather keep the actual functions simple.

rastervision_core/rastervision/core/data/label/semantic_segmentation_labels.py
AdeelH: Yes, but this is a little tricky. An elegant solution would make use of the existing `padding` param in `Box.get_windows()`. This would require modifying `Box.get_windows()` so that it applies padding to all sides instead of just the right and bottom as it currently does. Let me see if I can do that without breaking anything.

AdeelH: The current solution does not change anything in the training defaults. It just switches to symmetrical padding if `crop_sz` is not `None`. The user can choose to retrain using a different padding style if they feel unsatisfied with the predictions.

lewfish: +1

PR #1501, opened 2022-10-05
Update documentation site

### LINKED ISSUE #1240: Upgrade to Sphinx >= 4 and get doc styling to work
I had to downgrade Sphinx to 3.5 in order to get the docs styling to work. This seems to be due to the following change in 4.0:

"Support docutils-0.17. Please notice it changes the output of HTML builder. Some themes do not support it, and you need to update your custom CSS to upgrade it."

This issue discusses it: sphinx-doc/sphinx#9001 

Doing this upgrade isn't important, but I just wanted to note the issue.

https://github.com/azavea/raster-vision/pull/1239

### LINKED ISSUE #1507: TODO: update instructions for downloading the ISPRS Potsdam dataset in the docs
The ISPRS website has been updated and no longer requires filling out a form to request access to the dataset. The [instructions in the docs](https://docs.rastervision.io/en/0.13/examples.html#semantic-segmentation-isprs-potsdam) should be updated to reflect this.

The new download page is: https://www.isprs.org/education/benchmarks/UrbanSemLab/

AdeelH:
## Overview

**Demo: https://raster-vision--1501.org.readthedocs.build/en/1501/**

This PR updates the look and (to a lesser extent) structure of the documentation site. It does not make any significant changes to the content of the documentation.

**This PR also bumps up python to 3.8, GDAL to 3.2.2, and rasterio to 1.2**. See below for details.

Major changes:
- Use the [furo theme](https://github.com/pradyunsg/furo).
    - And remove custom CSS.
    - Another strong contender was `pydata_sphinx_theme` which is used by a lot of other libraries. It is currently not a good fit because it displays the top-level table of contents in the navigation bar instead of the sidebar and we have too many entries to properly fit in the navigation bar. We should consider switching to it in the future.
- Add a full, auto-generated API reference for all plugins, modules, classes, configs, functions.
    - This also includes source pages so you can see the actual code for each object.
    - This required some custom templates and sphinx extensions to get it to look right.
    - It also required defining `__all__`s in `__init__.py`s. Without this, `autosummary` gets tripped up by packages that have a module (file) of the same name as the package. E.g. when generating docs for `core.data.label_source` it would confuse it with `core.data.label_source.label_source`.
    - The older "Configuration API Reference" page has been removed.
- Add a "tutorials" gallery.
    - Generated from the example notebooks that were previously in `notebooks/` (now moved to `docs/tutorials/`).
    - Note that it is not required to execute notebooks when building docs (in fact, I have explicitly disabled this in `conf.py`). The html output saved as part of the notebook will be used to generate the doc pages instead.
- Update `sphinx` to `5.2.*`
    - This caused a conflict version with `flake8` over the correct `importlib-metadata` version. 
        - Fortunately, `flake8` only needed that dependency if using python < 3.8, so python was bumped to 3.8 in the `Dockerfile`.
            - This required `gdal` and `rasterio` (and apparently `imageio`) versions to be bumped. The combination `python 3.8, gdal 3.4.3, rasterio 1.3.*, imageio 2.22.*` seemed to cause some weird errors (segmentation fault when moving model to GPU, stalling when using num_workers > 0 on CPU), but `python 3.8, gdal 3.2.2, rasterio 1.2.*, imageio 2.22.*` works fine.

Minor changes:
- Constrain boto version requirement further to speed up version resolution.

### Screenshots
image

image

image

image

### Checklist

### Notes

- The global variables `registry` and `rvconfig` conflict with their respective module names causing the API documentation to not be generated for those modules. They (the variables) should be renamed to avoid the conflict.
- The `type_hint` field that we add to `Config`s should probably be renamed to `_type_hint` so that it is considered private and doesn't show up in the docs, since it is not usually of interest to users.
- A fortunate side-effect of moving notebooks under `docs/` will be to fix the repo's language stats. They are currently dominated by `.ipynb` files.

## Testing Instructions

- See https://raster-vision--1501.org.readthedocs.build/en/1501/

Closes #1240 
Closes #1507


AdeelH:
> * Move the Tutorials above the API reference in the ordering of sections. If you are scrolling through the docs home page, you will need to scroll through a huge number of lines for the API reference before you reach the Tutorials.

Done.

> * There is no syntax highlighting in the tutorial notebooks. Hopefully there's an easy to turn that on. There's highlighting on the source code linked from the API reference.

I had fixed it in #1506. Have moved that commit to this PR.

AdeelH:
Also moving the fix for #1507 to this PR.

lewfish (approved):

asoxien:
Hi adeel, lewfish

the #1501 throws me this error :
  **File "code/cowc_podsdam_fix.py", line 26, in <module>
    test: bool = False) -> ObjectDetectionConfig:
NameError: name 'ObjectDetectionConfig' is not defined**

Issue #1505, opened 2022-10-18
TODO: documentation improvements


AdeelH:
Follow-up to #1501 

**Major**
- [ ] Rewrite docs to emphasize the library aspect of RV (as opposed to the framework aspect).
- [ ] Rewrite README.
- [ ] Reorganize docs for better readability.
- [ ] Add more tutorial notebooks (#1459).

**Minor**
- [ ] Consider switching theme to `pydata_sphinx_theme`.
- [ ] Add dosctrings for modules.
- [ ] Ensure Google-style docstring format everywhere.
- [ ] Ensure docstrings and type hints for the most important classes/functions.
- [ ] Disambiguate `registry` and `rv_config` from their respective module names.
- [ ] Where feasible, replace `*args, **kwargs` with full list of arguments or link to the parent class's documentation.
- [ ] Consider using less automated API reference pages for modules. Example: https://github.com/scipy/scipy/blob/main/scipy/linalg/__init__.py,

lewfish:
What do you think about making it so that you can hit a button next to the notebooks in the docs and it will open it in Google Colab? We will have to modify the notebooks to have a preamble section where we install RV and any other necessary setup.

AdeelH:
I like that idea.

Here is an attempt:
https://github.com/azavea/raster-vision/pull/1506/commits/0e31373f5b5d9b7b21325466d557c42eb2c61261

Result:
https://raster-vision--1506.org.readthedocs.build/en/1506/tutorials/reading_labels.html

lewfish:
Looks good!

I noticed that when searching the docs for an argument to a Config class, the page I'm actually looking for is way down in the results. It seems to show links to source code before API docs. Maybe there's a way to specify a search path of sorts. 

For example, when searching https://docs.rastervision.io/en/latest/search.html?q=max_windows, I want it to give me this link first https://docs.rastervision.io/en/latest/api_reference/_generated/rastervision.pytorch_learner.learner_config.GeoDataWindowConfig.html

AdeelH:
[...]

Issue #1459, opened 2022-08-23
Add example notebooks


AdeelH:
As a follow-up to #1395, we should add Python notebooks that demonstrate how to use Raster Vision as a library.

If feasible, it would also be nice to automatically add rendered versions of these notebooks to the docs. See: https://docs.readthedocs.io/en/stable/guides/jupyter.html.

Some ideas:

### Basics
- [x] Reading raster data (#1470)
  - `RasterSource`, `MultiRasterSource`, `CRSTransformer`, `RasterTransformer`
- [x] Reading vector data (#1470)
  - `VectorSource`, `VectorTransformer`, `RasterizedSource`
- [x] Preparing labels (#1470)
  - `Labels`, `LabelSource`
- [x] Scenes and AoIs (#1506)
  - `Scene`
- [x] GeoDatasets (#1470)
  - `GeoDataset`, `SlidingWindowGeoDataset`, `RandomWindowGeoDataset`
  - different ways of instantiating
- [x] Training (#1506)
  - data augmentation
  - tensorboard
- [x] Making and saving predictions (#1506)
  - `Learner.predict_dataset()`, `LabelStore`
- [x] Evaluating predictions (#1506)
  - `Evaluator`
- [x] Visualizer (Lewis) (#1476)

### Other
- [ ] Full workflow for Potsdam segmentation that matches examples repo (Lewis)
- [x] Full workflow for ~Potsdam~ segmentation using RV + PyTorch Lightning (Lewis) (#1546)
- [ ] Reading GroundWork exports / reading STAC catalogs
- [x] Training on pre-chipped/non-georeferenced data (#1602)

lewfish:
Are you thinking that there will be different notebooks for the different checkbox points you listed? I was thinking we would have notebooks that covered whole workflows for specific datasets/tasks that would have the same scope as the examples on the Examples page that we already have, and in the course of these we would cover all these topics.

I was thinking that we might want to add an example of a lower-resolution land cover dataset since that's a common use case and isn't covered by the existing example which are all about cars and buildings and high-res imagery.

AdeelH:
>Are you thinking that there will be different notebooks for the different checkbox points you listed? I was thinking we would have notebooks that covered whole workflows for specific datasets/tasks

Yes, I'm hoping to have both. Shorter ones are easier to look up info in. I'm happy to take those on.

Also, this is a first draft. I think some of these can be merged together.

AdeelH:
[...]

PR #1506, opened 2022-10-21
Add more notebooks + misc. doc fixes and improvements

AdeelH:
## Overview

Notebook updates/addditions:
- Added a brief discussion of the `Labels` class to [reading_label.ipynb](https://raster-vision--1506.org.readthedocs.build/en/1506/tutorials/reading_labels.html#Labels)
- Added new notebook: [scenes_and_aois.ipynb](https://raster-vision--1506.org.readthedocs.build/en/1506/tutorials/scenes_and_aois.html)
- Added new notebook: [train.ipynb](https://raster-vision--1506.org.readthedocs.build/en/1506/tutorials/train.html)
- Added new notebook: [pred_and_eval.ipynb](https://raster-vision--1506.org.readthedocs.build/en/1506/tutorials/pred_and_eval_ss.html)
- ~Added an open-in-Colab button and installation instructions to all notebooks~ (See discussion in comments below).

Other changes:
- Problem with reference generation for `rastervision.pipeline.registry` and `rastervision.pipeline.rv_config` has been fixed by renaming the global `registry` and `rv_config` variables to `registry_` and `rv_config_` respectively.
- `rastervision2` plugin aliases have been removed.
- Fixed a problem with `LearnerPipelineConfig` not getting registered when `rastervision.pytorch_leaner` is the first RV plugin imported.
- Slightly refactored `Leaner.from_model_bundle`.
- Uploaded `LearnerPipelineConfig` bundles (in addition to the full pipeline bundles) to v0.13 model zoo.
    - Updated the `test.py` script to enable this.
- Made `test_ds` an optional input to `Learner`.
- Bumped `tensorboard` version.
- Added tests for `from_model_bundle` to the `Learner` unit tests.
- [One of the examples in reading_labels.ipynb](https://raster-vision--1506.org.readthedocs.build/en/1506/tutorials/reading_labels.html#Semantic-Segmentation---SemanticSegmentationLabelSource) used a non-public S3 URI to a Potsdam image before. Uploaded (only) this image to the public repo and updated the link.
- Enabled the use of `jinja` templates in all `.rst` files.
- Removed version hardcoding from docs.
- Updated version to `0.20-dev`.
- Misc. minor docs-related changes.

### Checklist

### Notes

N/A

## Testing Instructions

See https://raster-vision--1506.org.readthedocs.build/en/1506.

Makes progress on #1459 
Makes progress on #1505


lewfish:
I don't think this should be done in this PR or even in this release, but I would like to make it easier to use the prediction functionality without needing a model bundle or using the `Learner` class. I would like to extract out the prediction functionality from `Learner` (similar to what was done for `Visualizer`) and make it usable with an arbitrary PyTorch model (that has output of a specific form). This would make it easier to use RV's prediction functionality independent of the rest of RV. For example, if we used PyTorch lightning to train a model using RV's data reading functionality, it would be nice to take the model, and then generate predictions and evaluations with it.

lewfish:
> I don't think this should be done in this PR or even in this release, but I would like to make it easier to use the prediction functionality without needing a model bundle or using the `Learner` class (which requires pretty complicated configuration). I would like to extract out the prediction functionality from `Learner` (similar to what was done for `Visualizer`) and make it usable with an arbitrary PyTorch model (that has output of a specific form). This would make it easier to use RV's prediction functionality independent of the rest of RV. For example, if we used PyTorch lightning to train a model using RV's data reading functionality, it would be nice to take the model, and then generate predictions and evaluations with it.

I see we already discussed this a bit in https://github.com/azavea/raster-vision/issues/1459#issuecomment-1234416093. 
> You need to create a Learner with Learner(model=your_model, training=False)

I will try to do this. It still seems strange that you need to use a class called `Learner` just for making predictions.

rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
```diff
@@ -239,13 +248,14 @@ def main(self):
                 self.train()
                 if cfg.save_model_bundle:
                     self.save_model_bundle()
+        else:
+            self.load_checkpoint()
 
-        self.load_checkpoint()
+        self.stop_tensorboard()
+        self.sync_to_cloud()
```
lewfish, line 265: Shouldn't `sync_to_cloud()` should be called after `eval_model('valid')`?

README.md
```diff
@@ -27,7 +27,7 @@ There are several ways to setup Raster Vision:
     - We publish a new tag per merge into `master`, which is tagged with the first 7 characters of the commit hash. To use the latest version, pull the `latest` suffix, e.g. `raster-vision:pytorch-latest`. Git tags are also published, with the Github tag name as the Docker tag suffix.
 * Raster Vision can be installed directly using `pip install rastervision`. However, some of its dependencies will have to be installed manually.
 
-For more detailed instructions, see the [Setup docs](https://docs.rastervision.io/en/0.13/setup.html).
+For more detailed instructions, see the [Setup docs](https://docs.rastervision.io/en/0.20-dev/setup.html).
```
lewfish, line 30: Why is it 0.20-dev and not just 0.20?

AdeelH: I saw [another library doing it this way](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/__init__.py) and thought it made sense. Under this scheme, the version in the master branch would always be `<next planned version>-dev` and then the `-dev` suffix would be removed in the release branch.

Similar ideas are suggested in [this StackOverflow question](https://stackoverflow.com/questions/27185317/versioning-on-development-and-release-branches-git-flow).

rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
AdeelH: Yep, thanks for catching that. I believe it was only the `stop_tensorboard()` call I wanted to move up.

docs/conf.py
```diff
@@ -94,62 +151,83 @@
...
 
-MOCK_MODULES = ['pyproj', 'h5py', 'osgeo']
-sys.modules.update((mod_name, Mock()) for mod_name in MOCK_MODULES)
+            pip install rastervision
+    2.  Restart the notebook kernel.
+    3.  Set the ``GDAL_DATA`` environment variable required by ``rasterio``:
+
+        .. code-block:: shell
+
+            export GDAL_DATA=$(pip show rasterio | grep Location | awk '{print $NF"/rasterio/gdal_data/"}')
```
lewfish, line 174: To set env vars in a notebook you will need to do something like `%env MY_VAR=MY_VALUE`. Simply running `! export x=y` doesn't work because each invocation of `! ...` runs in its own shell (or something like that).

docs/conf.py
```diff
@@ -94,62 +151,83 @@
...
-    @classmethod
-    def __getattr__(cls, name):
-        return MagicMock()
+    1.  :ref:`Install Raster Vision <install raster vision>`:
 
+        .. code-block:: shell
 
-MOCK_MODULES = ['pyproj', 'h5py', 'osgeo']
-sys.modules.update((mod_name, Mock()) for mod_name in MOCK_MODULES)
+            pip install rastervision
```
lewfish, line 168: This will only work if the notebook code is compatible with the latest release of RV.

docs/conf.py
AdeelH: Good call

docs/conf.py
AdeelH: Since there does not seem to be a solution do this. I would suggest dropping the Colab button from this PR. What do you think?

lewfish: Yeah, makes sense.

AdeelH: Done. The colab-button code is preserved here for future reference: https://github.com/AdeelH/raster-vision/blob/docs2_w_colab/docs/conf.py#L154-L175.

lewfish (approved):

Issue #1247, opened 2021-08-17
Raster Vision Dependencies Falling Out-of-date


mablem8:
## 🐛 Bug

The dependencies as described by the Ubuntu 16.04 [Dockerfile](https://github.com/azavea/raster-vision/blob/4eaa709ea4beeddded048fcbf670306d42262d9d/Dockerfile#L8) are falling out-of-date. Native installation of Raster Vision is not possible with the latest version of Ubuntu (20.04) without compiling some dependencies from source.

## To Reproduce

Native installation on Ubuntu 16.04 can be achieved using the following commands:
```bash
sudo apt install software-properties-common python-software-properties
sudo apt install wget=1.*
sudo apt install git=1:2.*
sudo apt install python-protobuf=2.*
sudo apt install python3-tk=3.*
sudo apt install jq=1.5*
sudo apt install build-essential
sudo apt install libsqlite3-dev=3.11.*
sudo apt install zlib1g-dev=1:1.2.*
sudo apt install libopencv-dev=2.4.*
sudo apt install python-opencv=2.4.*
sudo apt install unzip
sudo apt install curl
if [ ! -d "$HOME/Downloads" ]
then
  mkdir $HOME/Downloads
fi
cd $HOME/Downloads
wget https://repo.anaconda.com/miniconda/Miniconda3-4.7.12.1-Linux-x86_64.sh
bash Miniconda3-4.7.12.1-Linux-x86_64.sh -b
source $HOME/miniconda3/bin/activate
conda init
conda config --set auto_activate_base false
conda update -n base -c defaults -y conda
conda deactivate
sudo shutdown -r now
conda create -n "rv" python=3.6.13 -y
conda activate rv
conda install -c conda-forge gdal=3.0.4 -y
python3 -m pip install rastervision==0.13
```

These steps cannot be completed successfully on Ubuntu 20.04. Specifically, the following are not available:

```
python-protobuf=2.*
jq=1.5*
libsqlite3-dev=3.11.*
libopencv-dev=2.4.*
python-opencv=2.4.*
```

Also, it appears that some of these dependencies are not actually needed to install and use Raster Vision. The [Setup](https://docs.rastervision.io/en/0.13/setup.html#installing-via-pip) documentation glosses over native installation and could be updated to specify which dependencies are required.

## Expected behavior

[...]

PR #1512, opened 2022-10-25
Upgrade environment and dependencies

lewfish:
## Overview

This PR updates the Dockerfile to use later versions of Ubuntu (20.04), CUDA (11.3), and Python (3.9). It also updates all Python dependencies (except for the `yapf` dev dependency) to their latest fully pinned version.

### Notes

* Updating `yapf` led to the `format_code` script making many changes across the codebase and some new `flake8` errors, and it didn't seem worth the trouble to update it right now since it's a development dependency.
* There are a few new deprecation warnings coming from `torch` after upgrading, but they require moderate changes to RV, and we have a few versions until we will be forced to update anything.
* There are some deprecation warnings coming from `geopandas` and `shapely` which appear to have been there even before these upgrades. These will need to be fixed before upgrading to shapely 2.0, and will require some moderate changes to RV. There is currently a preview release of shapely 2.0

## Testing

This was tested by:
* running the unit and integration tests on CI, my laptop (with an ARM64 version of the Dockerfile which will be added in a subsequent PR), and an EC2 instance with a GPU
* running 3 full examples (`spacenet-rio-cc`, `isprs-potsdam-ss`, `cowc-potsdam-od`) on Batch and getting similar output to before this PR
* checking that Jupyter notebooks still work locally

@AdeelH I wasn't able to check Jupyter on an AMD64 machine, so you should test that locally.

Connects #1247


Dockerfile
```diff
@@ -1,38 +1,40 @@
...
+# This is to prevent the following error when starting the container.
+# bash: /opt/conda/lib/libtinfo.so.6: no version information available (required by bash)
+# See https://askubuntu.com/questions/1354890/what-am-i-doing-wrong-in-conda
+RUN rm /opt/conda/lib/libtinfo.so.6 && \
+   ln -s /lib/x86_64-linux-gnu/libtinfo.so.6 /opt/conda/lib/libtinfo.so.6
 
 # needed for jupyter lab extensions
 RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash - && \
-    apt-get install -y nodejs
+   apt-get install -y nodejs
```
AdeelH, line 37: Tab size seems to have changed from 4 to 3.

docker/build
```diff
@@ -21,5 +21,5 @@ then
         exit
     fi
 
-    DOCKER_BUILDKIT=1 docker build --build-arg CUDA_VERSION="10.2" -t raster-vision-pytorch -f Dockerfile .
+     DOCKER_BUILDKIT=1 docker build --platform linux/amd64 --build-arg CUDA_VERSION="11.3.1" -t raster-vision-pytorch -f Dockerfile .
```
AdeelH, line 24: Tab size

rastervision_pipeline/rastervision/pipeline/__init__.py
```diff
@@ -2,6 +2,25 @@
 import logging
 import json
 
+# This is in a try block since the pytorch_learner plugin (which installs these
+# dependencies) may not be installed.
+try:
+    # torch is imported before anything else in RV in order to avoid a
+    # segmentation fault when calling model.to('cuda'). This was determined empirically,
```
AdeelH, line 9: Is it necessary for this to be in `pipeline` or can it also work if included before `import rastervision.pipeline` in `pytorch_learner`'s `__init__.py`?

rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
```diff
@@ -799,23 +799,19 @@ def _predict_dataloader(
...
-                x = self.to_device(x, 'cpu')
-                y = self.to_device(y, 'cpu') if y is not None else y
-                z = self.to_device(z, 'cpu')
-                if batched_output:
-                    yield x, y, z
-                else:
-                    for _x, _y, _z in zip(x, y, z):
-                        yield _x, _y, _z
-
-        self.model.train(model_train_state)
```
AdeelH, line 818: >We also get rid of the call to self.model.train because it will only get called if all the items are iterated through

This is unfortunate. Now this method ends up having a side-effect. Can you add a small note to the docstring that says that this will put the model in eval mode?

rastervision_pipeline/rastervision/pipeline/__init__.py
lewfish: I tried putting it in `core` and `pytorch_learner` and it didn't fix the problem there.

rastervision_pytorch_learner/rastervision/pytorch_learner/learner.py
lewfish: Yes

lewfish: Actually, this might be unnecessary since the `predict` method is getting called and we could just put the model into training mode at the end of that. That could have some performance implications though. I'm going to look into it.

lewfish: The model is put into train and eval mode in many other places in `Learner` and no attempt is made to put it back into the mode it was in before, so this change actually makes it more consistent with the rest of the class. This also mimics the behavior of PyTorch. There was some [discussion](https://discuss.pytorch.org/t/opinion-eval-should-be-a-context-manager/18998) about making `eval()` a context manager the same way `inference_mode` is. I'd rather just leave things as they are. I'm not opposed to changing the documentation to state that this is the behavior, but many methods will need to have this warning. 

Another issue is that the `predict` and `validate_step` methods never call `model.eval()` and are depending on the caller to do this which is dangerous. These methods should call `model.eval` themselves in case they are called directly.

I ran a little experiment and the time it takes to put a model into eval mode and then back into training mode is about 1/20 the time it takes to run a typical batch (of shape [32, 3, 256, 256]) through a Resnet18 when running on a Colab GPU. So the overhead of putting the model into eval mode for each batch is non-negligible, but not huge, and would make things safer.

AdeelH: Thanks for looking into this. I think doing it per batch is a bit much. PyTorch Lightning also seems [to do it only once at the start of training](https://github.com/Lightning-AI/lightning/blob/6b0b6b8903f3931f66360ac7ce0cd9e4d04e3f84/src/pytorch_lightning/trainer/trainer.py#L1154). We can leave it as it is for now.

Dockerfile
AdeelH: Other lines in this file are off too.

AdeelH (approved):

Issue #1549, opened 2022-11-12
Semantic Segmentation Experiment: Predicted labels are correct in `test_preds.png` but the actual predictions are wrong


theoway:
## To Reproduce

1. Use docker image rastervision-pytorch-v.0.13.1
1. Use the model_bundle.zip and rionegro_rgb_8_2.tif for `predict`. Files are available [here](https://drive.google.com/drive/folders/18kpE3y51LYlYzyborrE6oojNC4oFaSAt?usp=share_link).
1. This model was trained for prediction of cars and motorcyles. You should be getting correct predictions, but won't happen. 
You will get incorrect results similar to `labels.tif` file that is [here](https://drive.google.com/drive/folders/18kpE3y51LYlYzyborrE6oojNC4oFaSAt?usp=share_link).
Also, there should be vector outputs as well in the output folder that you provide. But that doesn't happen either.

The model had been trained using AWS batch. Here's the `get_config()` code:
```python
import os

from rastervision.core.rv_pipeline import *
from rastervision.core.backend import *
from rastervision.core.data import *
from rastervision.pytorch_backend import *
from rastervision.pytorch_learner import *

IDS = ['El_Retiro_1_3', 'El_Retiro_1_4', 'El_Retiro_2_1', ...]

def build_scene(uri, id, channel_order=None):
    label_file = id + '.geojson'
    label_uri = uri + label_file
    print("Label uri: ", label_uri)

    image_file = id + '.tif'
    image_uri = uri + image_file
    print("Image uri: ", image_uri)

    raster_source = RasterioSourceConfig(
        uris=[image_uri],
        channel_order=channel_order,
        transformers=[StatsTransformerConfig()])
    
    vector_source = GeoJSONVectorSourceConfig(
        uri=label_uri,
        default_class_id=0,
        ignore_crs_field=True)

    label_source = SemanticSegmentationLabelSourceConfig(
        raster_source=RasterizedSourceConfig(
            vector_source=vector_source,
            rasterizer_config=RasterizerConfig(background_class_id=0)))
    
    label_store = SemanticSegmentationLabelStoreConfig(
        vector_output=[PolygonVectorOutputConfig(class_id=1),
                        PolygonVectorOutputConfig(class_id=2),
                        PolygonVectorOutputConfig(class_id=3)])

    return SceneConfig(
        id=id,
        raster_source=raster_source,
        label_source=label_source,
        label_store=label_store)

def get_config(runner, raw_uri, root_uri, test=True):
[...]

Issue #1496, opened 2022-10-03
Reconcile chip-sampling functionality in `RVPipeline`s with `GeoDataset`s


AdeelH:
## 🚀 Feature

Chip-sampling via the `chip` command should be identical to chip-sampling via `GeoDataset`s.

## Motivation

There is some special chip-sampling logic implemented in `RVPipeline` subclasses that is not currently supported in the corresponding `GeoDataset` subclasses and vice versa, meaning chip-sampling via the `chip` command is not identical to chip-sampling via `GeoDataset`s.

## Pitch

1. Bring over any missing functionality from `RVPipeline`s to `GeoDataset`s.
2. Remove chip-sampling logic from `RVPipeline`s.
3. Move `chip` implementation to the `Backend` (same as `train` and `predict`).
    - This could also improve performance via the use of PyTroch `DataLoader`s.

## Alternatives

Just do (1) and take on the technical debt of maintaining duplicated code.

## Additional context

This will also involve reconciling the `*ChipOptions` classes with `GeoDataWindowConfig`.

## Notes
Functionality present in `RVPipeline`s but missing in `GeoDataset`s:
- chip classification
  - `chip_nodata_threshold`
- semantic segmentation
  - `chip_nodata_threshold`
  - filter chips with all null class pixels
  - `negative_survival_prob`, `target_class_ids`, `target_count_threshold`
- object detection
  - `label_buffer` -- doesn't seem particularly useful
  - the random positive chip sampling behavior is also a little different

Possible plan of work:
- move `GeoDataWindowConfig` to `rastervision.core`, rename to `ChipOptions`
  - make other `*ChipOptions` derive from it
- `GeoDataConfig` now takes this new `ChipOptions`
- `ChipOptions` must be specified in the `RVPipelineConfig` if the `chip` command is to be used.
  - it will be passed to the pytorch backend and used to instantiate a GeoDataset

PR #1556, opened 2022-11-15
Bug fix: do not zero out null class pixels

AdeelH:
## Overview

This PR removes the unintuitive and unnecessary behavior of filling in zeros for null class pixels while creating training chips for semantic segmentation.

This behavior is also clearly incompatible with the use of the background class as the null class.

See discussion in #1549.

### Checklist

### Notes

This also makes the chip sampling behavior more similar to `GeoDataset` behavior, and thus, connects to #1496.

### DIFF
--- a/rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation.py
+++ b/rastervision_core/rastervision/core/rv_pipeline/semantic_segmentation.py
@@ -5,8 +5,7 @@ import numpy as np
 
 from rastervision.core.box import Box
 from rastervision.core.rv_pipeline.rv_pipeline import RVPipeline
-from rastervision.core.rv_pipeline.utils import (fill_no_data,
-                                                 nodata_below_threshold)
+from rastervision.core.rv_pipeline.utils import nodata_below_threshold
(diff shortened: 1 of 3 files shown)


AdeelH:
@lewfish

lewfish:
I think the intention of the original behavior was to handle the case where part of an image is unlabeled, and is labeled with the null class in those unlabeled parts. For example, if an image of a house is unlabeled, then you don't want to teach it to associate houses with the null class. It's better to just zero it out, and then the model will learn to associated zero pixels with the null class. Does this fix still support that use case?

lewfish:
It should be possible to use the AOI functionality to handle this use case, so we don't have to preserve this behavior. I think a long time we were given a dataset when null labels were used to indicate "not labeled".

AdeelH:
Yes, AOIs are the idiomatic way to handle this. 

The only case when we would still want to zero out pixels is when the AOI is too irregular and you're using `Box.filter_by_aoi(within=False)`. In that case you might want to zero parts of the window that overflow the AOI. But let's worry about that edge case when/if we ever come across it.

PR #1672, opened 2023-01-20
Add error handling for empty DataLoader

perliedman:
## Overview

This adds two small error checks for when the `DataLoader` returns an empty iterator. Without this, code crashes with hard-to-understand internal errors.

### Checklist

### Notes

## Testing Instructions

Learner:

* Create a dataset which is empty (for example, a random window sampler with an AOI that is impossible to hit)
* Run `train_epoch`
* `ValueError` should be thrown (previosly resulted in an `IndexError` when `outputs` was accessed)

Visualizer:

* Create a dataset which is empty (for example, a random window sampler with an AOI that is impossible to hit)
* Call `plot_batch`
* `ValueError` should be thrown (previously resulted in an `NameError` since `x` wasn't defined)

### DIFF
--- a/rastervision_pytorch_learner/rastervision/pytorch_learner/dataset/visualizer/visualizer.py
+++ b/rastervision_pytorch_learner/rastervision/pytorch_learner/dataset/visualizer/visualizer.py
@@ -153,8 +153,11 @@ class Visualizer(ABC):
         """
         collate_fn = self.get_collate_fn()
         dl = DataLoader(dataset, batch_sz, collate_fn=collate_fn, **kwargs)
-        for x, y in dl:
-            break
+        try:
+            x, y = next(iter(dl))
+        except StopIteration:
+            raise ValueError('dataset did not return a batch')
+
         return x, y
 
     def get_plot_nrows(self, **kwargs) -> int:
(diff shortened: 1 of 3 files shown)


AdeelH:
Thank you for the PR! 

This looks good to me, but I'll let @lewfish take a look in case the `for x, y in dl: break` construction was important for some reason.

I think a unit test for `Learner.train_epoch()` might be tricky/too involved, but a unit test for `Visualizer.get_batch()` would be nice.

rastervision_pytorch_learner/rastervision/pytorch_learner/dataset/visualizer/visualizer.py
```diff
@@ -153,8 +153,11 @@ def get_batch(self, dataset: 'Dataset', batch_sz: int = 4,
         """
         collate_fn = self.get_collate_fn()
         dl = DataLoader(dataset, batch_sz, collate_fn=collate_fn, **kwargs)
-        for x, y in dl:
-            break
+        try:
+            x, y = next(iter(dl))
```
lewfish, line 157: LGTM. I forgot I could just use `next(iter(dl))`

lewfish:
Thanks for the contribution. Even though it's a very small change we'll need for you to sign a CLA before merging. Sorry for the trouble. See https://github.com/azavea/raster-vision#contributing

perliedman:
@lewfish :+1: no worries, I e-mailed you the license agreement just now!

AdeelH (approved):
Looks good! Thank you!

PR #1674, opened 2023-01-24
Only set default stride if stride value is missing in `GeoDataWindowConfig`

perliedman:
## Overview

Yet another tiny fix for what I believe is a bug: when passing `stride` to `GeoDataWindowConfig`, it is currently always overwritten with the default value (`size`).

With this PR, the default is only set if `stride` has not been explicitly set, much like the logic for `size_lims` in the same class.

### Checklist

## Testing Instructions

* Added a unit test for both the default (no explicit `stride` parameter, value should fallback to `size`) and when `stride` is set, so running unit tests should be enough to verify that this works
* Creating a dataset with a non-default stride and check that the correct number of windows are created


AdeelH (approved):
Good catch! That definitely looks like a bug. Thank you for the fix!

Issue #1130, opened 2021-03-09
Automate publishing packages to PyPI


ColeKettler:
Currently we're manually publishing our [various Raster Vision packages](https://pypi.org/search/?q=rastervision) to the `azavea` namespace on PyPI. We should automate this process to be part of our CI release workflows, like we've done with [`django-ecsmanage`](https://github.com/azavea/django-ecsmanage/blob/develop/.github/workflows/release.yml).

Depends on #1129.

AdeelH:
Do we have experience publishing packages to conda? Can we also include the automation of conda publication to this issue?

PR #1915, opened 2023-09-08
Add scripts for building packages and publishing them to PyPi

AdeelH:
## Overview

This PR adds scripts to build and publish packages to PyPi, thus automating a manual and tedious part of the release process. It also updates the release instructions accordingly. 

This PR is a step towards #1130.

### Checklist

## Testing Instructions

* These scripts were used to publish v0.21.1.
* Check release process doc in CI docs build.

### DIFF
--- a/scripts/pypi_build
+++ b/scripts/pypi_build
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+SOURCE="${BASH_SOURCE[0]}"
+while [ -h "$SOURCE" ] ; do SOURCE="$(readlink "$SOURCE")"; done
+SCRIPTS_DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+SRC_DIR="$( cd -P "$( dirname "$SCRIPTS_DIR" )" && pwd )"
+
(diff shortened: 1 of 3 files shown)


jamesmcclain:
@AdeelH 👀

scripts/pypi_build
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
jamesmcclain, line 1: I don't see anything that looks obviously wrong to me ... if the test publication works as expected, than it should be good.

jamesmcclain (approved):
Assuming the test publication works as expected, this looks good to me.

scripts/pypi_publish
```diff
@@ -0,0 +1,43 @@
...
+    fi
+    echo "Done."
+    cd "$SRC_DIR"
+}
+
+# Publish each plugin
+for plugin in "${plugins[@]}"; do
+    publish_plugin "$plugin"
+done
+
```
jamesmcclain, line 35: The only thing that I might suggest -- from a belt-and-suspenders point of view -- is to force a test publication before a real one.  If the person says passes `--test` then it will terminate after the test publication, if they do not then it will publish to testpypi then prompt them to inspect the published artifacts before proceeding with the real publication.

AdeelH: To keep things simple, I've made it so that
- `--test` works as before
- With no `--test`, you have to go through a `Actually publish to PyPi? (y/N)` prompt
- `-y` bypasses the prompt (for when/if we choose to fully automate things)

PR #1921, opened 2023-09-17
Save model weights for each epoch 1720

### LINKED ISSUE #1720: Save model's weights for each epoch
## 🚀 Feature

Currently, during training, the weights are saved after every epoch to `last-model.pth` in the folder `training` (i.e. the file is overwritten after every epoch). Once the training is complete, a bundle is created in the folder `bundle` and `last-model.pth` file is copied to the bundle as `model.pth`.

However, it would be interesting to keep the weights of the model for **all the epochs**.  
If we define the "best" model as the model for which a chosen metric was the highest (like map50 for object detection, on the validation set), this is not necessarily the "last" model.  

## Motivation
This is related to the discussion #1710 (model fine-tuning for an object detection task)  

## Pitch
2 propositions : 
* keep the weights at each epoch, in order to select the best model at the end of the training. 
* keep the weights of the best model only (based on some metric), even if the file is overwritten after every epoch (but that would mean add the relevant metric as a parameter)

## Alternatives

So far, the only alternative is to manually kill the job in the middle of training, if we monitor a decrease in the metric. In that case, the weights from the last completed epoch will be in `last-model.pth`.

## Additional context
This being said, as [AdeelH](https://github.com/AdeelH) wrote in #1710, one should be careful about how much stock to put in the model's performance on the validation set. If your validation set is too small or not representative of the test set or the data in production, then you run the risk of overfitting the validation set.

mmcs-work:
## Overview

Save model weights for each epoch.

### Notes
From #1720, proposition 1 is followed. Last checkpoint is still kept as `last-model.pth`. Instead of overwriting, it copies the previous checkpoint to `model-ckpt-epoch-{N}.pth` where `N` is the epoch number.

## Testing Instructions

* From either running the notebooks or the unittests where `Learner` is extended. The option `save_all_checkpoints` can be added as in
```
learner = SemanticSegmentationLearner(
    cfg=learner_cfg,
    output_dir='./train-demo/',
    model=model,
    train_ds=train_ds,
    valid_ds=val_ds,
    save_all_checkpoints=True
)
```
Closes #1720

### DIFF
--- a/rastervision_pytorch_backend/rastervision/pytorch_backend/pytorch_chip_classification_config.py
+++ b/rastervision_pytorch_backend/rastervision/pytorch_backend/pytorch_chip_classification_config.py
@@ -53,7 +53,8 @@ class PyTorchChipClassificationConfig(PyTorchLearnerBackendConfig):
             test_mode=self.test_mode,
             output_uri=pipeline.train_uri,
             log_tensorboard=self.log_tensorboard,
-            run_tensorboard=self.run_tensorboard)
+            run_tensorboard=self.run_tensorboard,
+            save_all_checkpoints=self.save_all_checkpoints)
         learner.update()
         learner.validate_config()
         return learner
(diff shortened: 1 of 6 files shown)


AdeelH:
Looks like there are linter errors. You can run `scripts/style_tests` locally to find those. And also `scripts/format_code` to format your code correctly.

If you are using VS Code, you can install the `flake8` extension and set `yapf` as the formatter to ensure that your code is always compliant. Be sure to install the `flake8` and `yapf` versions from `requirements-dev.txt`.

AdeelH (approved):
This looks good to me. Thank you!

mmcs-work:
> This looks good to me. Thank you!

Thanks for your help.

PR #2324, opened 2025-09-28
Fix GitHub license detection

adamjstewart:
## Overview

At the moment, GitHub is unable to detect the license of this project. This is because `LICENSE` contains the Apach-2.0 file header, not the full Apache-2.0 license.

### Checklist

- [ ] Added unit tests, if applicable
- [ ] Updated documentation, if applicable
- [ ] Added `needs-backport` label if the change should be back-ported to the previous release

### Notes

See https://github.com/licensee/licensee/issues/103 for details.

## Testing Instructions

Can be tested using https://github.com/licensee/licensee

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright [2025] [Element 84, Inc.]
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/LICENSE
+++ b/LICENSE
@@ -1,13 +1,202 @@
-Copyright 2017 Azavea [http://www.azavea.com]
 
-Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,15 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
mentzerk: This should be 2025 Element 84, Inc.

AdeelH (approved):
Thanks, @adamjstewart!

LICENSE
adamjstewart: Technically it shouldn't, this is only when you copy the header into other files. But it seems like detection is working now, so that's all that matters to me.

